PR #31, opened 2017-07-05
Layout issues with the widget editor

clementprdhomme:
This PR fixes two issues with the layout of the widget editors:
- Several columns could be displayed on the same lines
- Columns with long names would break the layout if placed in a box container

The solution is to use overflow everywhere. It doesn't reduce the usability but that's not an ideal solution either. Here are the list of issues that haven't been addressed:
- To see the full name of the columns the user might be forced to scroll horizontally
- To access the aggregation functions, the user might have to scroll horizontally

I've created [a task on Pivotal](https://www.pivotaltracker.com/story/show/148330129) so Dani can review the design and maybe give us some solutions to the remaining issues.

[Pivotal task](https://www.pivotaltracker.com/story/show/148324491)

### DIFF
--- a/css/components/widgets/dimensions_container.scss
+++ b/css/components/widgets/dimensions_container.scss
@@ -1,6 +1,12 @@
 .c-dimensions-container {
   .c-column-container {
     margin-bottom: $space-1;
+    // Here we prevent the columns from growing too much
+    // This isn't an ideal solution at all, but that's the
+    // easiest that doesn't reduce the usability of the tool
+    // (i.e. the user can still see the full name of the column)
+    max-width: 172px;
+    overflow-x: scroll;
     &:last-child { margin-bottom: 0; }
   }
 }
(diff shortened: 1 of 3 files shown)


clementprdhomme:
@pablopareja I need more info: which browser, which OS?

clementprdhomme:
Can't reproduce on Safari, Chrome or Firefox. Can you tell me where you've found the issue (explore detail or CMS) and on which browser and OS?

<p align="center">
Screenshot of the widget editor with no apparent layout issues
</p>

pablopareja:
I was using Chrome Version 59.0.3071.115 on Mac.
The screenshot was taken on the Explore Detail page

clementprdhomme:
Can you try again restarting your local server? I've tried on the same version of Chrome and I don't have the issue.

If you still have the problem, do you mind if you share your screen with me so we can see why it's not working on your computer?

pablopareja:
screen shot 2017-07-09 at 5 42 06 pm

I keep getting some issues with the dimensions containers...
Anyway let's talk about this on Monday but we should first check if the new design from Dani is confirmed. In that case those new styles may fix the whole thing

clementprdhomme:
This PR is outdated.

PR #37, opened 2017-07-06
Fix the wrong values of the tooltip

clementprdhomme:
<p align="center">
Pie chart with a tooltip containing values
</p>

This PR fixes issues with the values of the tooltip:
- It wouldn't work properly for the pie charts because it was expecting an x axis
- It would keep the previous value when leaving a mark
- The algorithm would delete any eventual signals
- The code has been refactored

[Pivotal task](https://www.pivotaltracker.com/story/show/148345887)


clementprdhomme:
@pablopareja Just rebased on `develop`. Before doing so, I also had an issue running `npm install` after installing the modules. Now it seems to be fixed.

PR #124, opened 2017-07-24
Bye Annoying Sticky Tooltip ™, Hello Sexy Tooltip ™

clementprdhomme:
This PR makes the tooltip of the charts even more sexier than ever before.

<p align="center">
Tooltip shown on top of a scatter plot
</p>

**Changes:**
* Dates are now formatted as the ticks of the charts
* Labels are associated with the values **[1]**
* Brand new design for the tooltips **[2]**
* Slight performance improvement
* No more leaking tooltips outside of the chart
* Fixed an issue with the aggregation functions **[3]**
* Tooltips more shy with the scatter plot **[4]**
* Updated [documentation](https://docs.google.com/a/vizzuality.com/document/d/1t_gUgb3ti_LUmQy3Gl3cRy6sLbwlyzqT3GH_6E3-bak/edit?usp=sharing)

**[1]** The labels are the name of the columns used to draw the chart + the aggregate function, if any. Older widgets might not display the labels because they need to be updated (see the [Pivotal task](https://www.pivotaltracker.com/story/show/149428295)).

**[2]** The tooltip component has been updated to the [latest design](https://projects.invisionapp.com/d/main#/console/11337456/240485596/preview). In order to do so, a tip has been added. By default, the tip is added to the bottom of the tooltip and is resetted to this position each time it's hidden. **Because of this, the layout of other pages might be affected.**

**[3]** If "none" would be selected, the store would save `none` instead of `null` (the default state).

**[4]** Unlike the bar or line chart, the scatter plot can have multiple values for the same horizontal position. Because of that, we can't determine which value to show in the tooltip when the cursor is not positioned on top of a specific point. For the scatter chart, the tooltip will be then only displayed when the dot is hovered.


clementprdhomme:
**Please don't merge.**

There's an issue with the values displayed with this dataset: https://staging.resourcewatch.org/data/explore/6d99441e-5faa-4c61-967f-01c9fe60624b

mbarrenechea:
Also, it sometimes doesn't get the correct position.

captura de pantalla 2017-07-25 a las 9 42 45 2
captura de pantalla 2017-07-25 a las 9 43 04 2
captura de pantalla 2017-07-25 a las 9 46 46

mbarrenechea:
Maybe the problem is that you are sorting the `bisectDate` but you are not showing it sorted...

clementprdhomme:
Hey @pablopareja, @mbarrenechea!

Thank you for you help.

I think I've solved all the issues you mentioned. I also made a major performance boost to the tooltip (x2) by adding a few `shouldComponentUpdate` to some components. You can see the changes for that [here](https://github.com/resource-watch/resource-watch/pull/124/commits/18298d9fbbddf6023d3060714a3a8ee336747f06).

Let me know if this can be merged.

clementprdhomme:
There's an issue with the tooltip not showing in the header. I'm on it.

PR #141, opened 2017-08-04
Feature/new dashboards

clementprdhomme:
This PR updates the dashboard pages so they can take into account the dashboards created through the wysiwyg.

I've created a sample dashboard to test the styles: http://localhost:3000/data/dashboards/cities
There you can see how landscape and portrait images, lists, basic formatting (bold, italic, underlined) and the widgets would look like.

By default, all the images will resize to take the full width, if landscape, or won't be taller than the screen height, if portrait. In order to have a big widget or to have two of them side by side, you need to use a wrapper with the class `columns` and then for each column, you'll have a div with the class `column-X` (`X` being `1` or `2`). For example, this would display two columns side by side:
```html
<div class="columns">
  <div class="column-2">Column 1</div>
  <div class="column-2">Column 2</div>
</div>
```

And this would display one column taking the full width:
```html
<div class="columns">
  <div class="column-1">Some important content here</div>
</div>
```

**NOTE:** I've been unable to upload a photo for the dashboard through the API, but I've locally tested displaying a base64 image for the dashboards list.

### DIFF
--- a/css/components/app/pages/dashboards.scss
+++ b/css/components/app/pages/dashboards.scss
@@ -126,6 +126,60 @@
 
   .user-content {
     margin-top: 30px;
+
+    img {
+      max-width: 100%;
+      border-radius: 4px;
+
+      @media screen and (min-width: $grid-row-width) {
+        max-width: $grid-row-width;
+        max-height: 100vh; // Improvement for portrait images
+        margin: 0 auto; // Needed for portrait images
+        // The following 3 lines are used to center the photo
+        // with the content
+        position: relative;
+        left: 50%;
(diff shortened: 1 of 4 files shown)


davidsingal:
Could we use other class names? It could have some conflicts with grid style class names (row column columns). Maybe something like `dashboard-row` and `dashboard-column-(1|2|3)` or another shorter name.

clementprdhomme:
I've renames the classes as following:
* `.columns` ➡️ `.dashboard-row`
* `.column-X` (with `X` being `1` or `2`) ➡️ `.dashboard-column-X`

PR #167, opened 2017-08-22
Save raster and map widgets

clementprdhomme:
This PR lets the user save raster (maps and charts) as well as standard maps as widgets. In addition, the user has the possibility to change the type of visualisation of a created widget. For example, a chart can be converted to a map.

Other changes include:
* the embed option of the widget editor for the maps has been moved next to the zoom buttons
* the styles and behaviour of the widgets in My RW have changed
* the icons have been updated through iconmoon


clementprdhomme:
Hey Pablo,

Thank you for the review. I think I can fix: **1.** (regression), **3.** (new bug **☨**), **5.** (probably new bug) and **6.** (regression).

Points **2.** and **4.** are previous bugs that are already in `develop` so if you don't mind, I'd rather fix them later so we can merge the new architecture right away. Could you add them to Pivotal?

**(☨)** Regarding **3.**, you probably copied the local address so it will never work. You need to open the link in a new tab and you should be able to see the widget. Nevertheless, I just tried and noticed the map is not displayed (the dataset info is though).

clementprdhomme:
Hi @pablopareja,

Unlike what I previously said, I've fixed the 6 points you mentioned above. Could you review again?

Thanks.

pablopareja:
Hey @clementprdhomme,

I'm getting issues when trying to save a map widget from MyRW
image

clementprdhomme:
Hey Pablo,

Can you try to delete the widget and re-create it? I think it might be because the widget was created at a time where we didn't store the type of visualisation in the API (chart, raster chart or map).

pablopareja:
I was creating a new widget...

pablopareja:
I've just tried again and I'm still getting the same issue

image

pablopareja:
There's an issue when trying to create a widget with a pre-existing title, the form interface disappears:
_(both in the case of a map and a normal widget)_

image

image

pablopareja:
One more thing, map widgets created from MyRW don't appear in `My Widgets` for some reason

PR #178, opened 2017-08-30
Explore - filters resume

andresgnlez:
image

Implements filters resume component. They need some styling, I just added a task for the designer (https://www.pivotaltracker.com/story/show/150666488).

Also, we are now displaying the values (not the labels) of these filters, shouldn't better display the labels? This would require some changes though.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+indent_style = tab
+indent_size = 4
+trim_trailing_whitespace = false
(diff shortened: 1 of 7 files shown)


andresgnlez:
`About the values shown in the list component, yes we should show labels instead of the IDs since in some cases the ids can look quite ugly.
How difficult would it be to make that change?`

We could use the local state to achieve this. Shouldn't be very problematic.

`Topics filters are not being applied in the case where there're no geographies or data types selected. The request however is made and it returns data.`

Didn't know this condition. What's the point of enabling the selector if it depends on the geographic selector? By now, I will just skip the request if the condition doesn't occur. But I found pointless allowing the user to select a topic if this one is not going to work unless a geography is selected.

pablopareja:
Regarding the point 3. I see that I didn't explain myself right, what I meant is that that is *unexpected behavior* and it should be fixed so that the topics selector works in the same way as the other two 😁

pablopareja:
There still are two things missing:

1. Selected elements lists should all occupy the least space possible, that's to say, instead of using different _rows_ for each type of elements, the same row should be divided in 1, 2, or 3 columns depending on the selection made by the user.
image

2. Selected concepts shown in the list should have a cross icon in the right side that could be used as a shortcut to deselect them without interacting with the trees.

pablopareja:
The changes mentioned in the last comment will be done in another pull request. 
Merging this...

PR #191, opened 2017-09-04
Add metadata title and filter for datasets

ivanyukio:
When displaying datasets, always show the metadata title if present, otherwise show dataset title.

Filtering datasets will always look in the metadata title if it's present, otherwise it'll look in the dataset title.


components/app/explore/DatasetWidget.js
```diff
@@ -133,7 +134,7 @@ class DatasetWidget extends React.Component {
                 route={'explore_detail'}
                 params={{ id: this.props.dataset.id }}
               >
-                <a>{dataset.name}</a>
+                <a>{metadata == undefined ? dataset.name : metadata.attributes.name}</a>
```
pablopareja: The metadata name that should be used is:
metadata.attributes.info.name
image

Also:
1. please use `!metadata` instead of `metadata == undefined`
2. please always use `===` instead of `==`

pages/app/ExploreDetail.js
```diff
@@ -212,7 +212,7 @@ class ExploreDetail extends Page {
                 />
 
                 <h1>
-                  { dataset && dataset.attributes && dataset.attributes.name}
+                  {metadataAttributes == undefined ? (dataset && dataset.attributes && dataset.attributes.name) : metadataAttributes.name}
```
pablopareja: Same as the comment above.

selectors/explore/filterDatasets.js
```diff
@@ -14,8 +14,15 @@ const getFilteredDatasets = (_list, _filters) => {
     let conceptsCheckPassed = true;
 
     if (search && search.key === 'name') {
-      if (it.attributes.name.toLowerCase().match(search.value.toLowerCase())) {
-        searchFilterPassed = true;
+
+      if (it.attributes.metadata[0]) {
+        if (it.attributes.metadata[0].attributes.name.toLowerCase().match(search.value.toLowerCase())) {
```
pablopareja: Use the name under `info`

ivanyukio:
@pablopareja I changed everything you asked, please let me know if it's good, thanks!

PR #201, opened 2017-09-07
Use metadata title for dataset when possible

ivanyukio:
@pablopareja There's still a place using the dataset title in the datasets tab under My RW, but I couldn't add a metadata to my datasets due to a 403 on the request.


components/app/myrw/widgets/pages/new.js
```diff
@@ -184,29 +184,29 @@ class WidgetsNew extends React.Component {
...
+      { filters: { userId: this.props.user.id }, includes: "metadata" }).then((response) => {
       this.setState({
         datasets: [...this.state.datasets, ...response.map(dataset => ({
           id: dataset.id,
           type: dataset.type,
           provider: dataset.provider,
           tableName: dataset.tableName,
-          label: dataset.name,
+          label: dataset.metadata[0] && dataset.metadata[0].attributes.info ? dataset.metadata[0].attributes.info.name : dataset.name,
           value: dataset.id
```
pablopareja: Please fix Eslint issues such as use of double quote or lines that are too long. 
Also please use in this kind of cases something like:
```
const metadata = dataset.metadata[0];
label: metadata && metadata.attributes.info ? ....
```
so that you don't end up having huge lines

ivanyukio:
The DatasetsListCard is the one that I commented where I couldn't test since I was not able to create a metadata for my dataset. But I'll change it and follow the same structure as the other ones as well as apply the other requested changes, thanks!

ivanyukio:
Made the changes, please review again, thanks!

PR #210, opened 2017-09-11
Revamps explore selectors

andresgnlez:
Related with: https://www.pivotaltracker.com/story/show/150880146

image


davidsingal:
Could you review the margin between elements? I think it should be shorter.

PR #234, opened 2017-09-19
Feature/layers opacity

bitTal:


bitTal:
2.- Opacity changes when the dot is released, so the done button is to close the tooltip
3.- Buttons are aligned that way in the design

pablopareja:
I'm getting a new error thrown every time I change the value of the slider:
image

PR #236, opened 2017-09-19
Feature/raster improvements

clementprdhomme:
This PR updates the queries used to display the data of the raster datasets and specifically the queries for the categorical bands.

To test the support for Carto datasets, use [this one](http://localhost:3000/data/explore/ad790c87-fe9e-4405-891d-de7c2ddfda79) and to test the GEE ones, you can use [this one](http://localhost:3000/data/explore/8ecf07f5-3cae-4275-adfe-918d04439a1a). Both examples only contain categorical bands.

The graphs won't show for the GEE datasets because the query that fetches the data is not currently supported by the API. As soon as @aagm implements a fix, it should work properly (the code is based on a mockup result) but further testing should be done anyway.

I'll update this PR once I hear back from Alicia.

### DIFF
--- a/utils/widgets/WidgetHelper.js
+++ b/utils/widgets/WidgetHelper.js
@@ -209,43 +209,6 @@ export function getChartInfo(dataset, datasetType, datasetProvider, widgetEditor
 }
 
 /**
- * Return the number of bins for the raster
- * @export
- * @param {string} dataset - Dataset ID
- * @param {string} tableName - Name of the table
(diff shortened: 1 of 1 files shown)


clementprdhomme:
The PR has been updated with the new query from Alicia.

Note: categorical GEE bands can have a "No data" column.
<p align="center">
Bar chart with one bar labelled as 'no data'
</p>

PR #245, opened 2017-09-21
Fixes displaying deleted resources after removing them

andresgnlez:
https://www.pivotaltracker.com/story/show/151256369


mbarrenechea:
We need to change all of them. I mean Pages, Dashboards, Partners

PR #258, opened 2017-09-25
Feature/metadata multiple sources

andresgnlez:
TO-DO: render list of sources in the main form.


components/admin/metadata/form/SourcesContentModal.js
```diff
@@ -0,0 +1,103 @@
...
+
+// components
+import ContentGroup from 'components/ui/ContentGroup';
+import Source from 'components/admin/metadata/form/Source';
+import { SOURCE_ELEMENTS } from 'components/admin/metadata/form/constants';
+
+class SourcesContentModal extends React.Component {
+
+  componentWillMount() {
+    const { sources, setTmpSources } = this.props;
```
mbarrenechea: Try a dataset that doesn't have any source added. Could you please set the initial state as an array with one empty object? Just the same you are doing when you try to remove all sources...

image

mbarrenechea: Also, could you list the selected sources in the metadata form? At least string separated by commas

andresgnlez: image

Added source list.

andresgnlez:
I've removed input validation for the time being.. This is giving us more problems than solutions in some cases.

PR #264, opened 2017-09-26
Display the aliases of the fields in the editor whenever possible

clementprdhomme:
This PR displays the aliases of the fields when present in the widget editor. The aliases can be seen for the "Chart" visualisation type and for the table as well.

You can test the feature with this dataset: `f2fe7588-6d1b-400e-b79c-0c86bf1273ea `.


clementprdhomme:
Sorry I completely forgot about that. I've just updated the PR.

PR #284, opened 2017-10-03
Fix the distorted charts in My RW > My widgets

clementprdhomme:
The charts would be forced to take their container's full width and height whereas the bar and pie charts determine their sizes by themselves. This would distort them.

[Pivotal task](https://www.pivotaltracker.com/story/show/151472665)

### DIFF
--- a/css/components/widgets/widget_card.scss
+++ b/css/components/widgets/widget_card.scss
@@ -36,7 +36,13 @@
         display: flex;
 
         canvas {
-          width: 100%;
+          // NOTE:
+          // Both the width and the height can't be defined because
+          // the "self-sizing" charts will be distorted (in My RW >
+          // My widgets)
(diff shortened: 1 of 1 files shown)


clementprdhomme:
@pablopareja The grid view shows the charts as expected. The problem you see is probably because the widgets are bar charts and their thumbnails are broken. You can try to scroll within the widgets to see the bars or create widgets based on pie, line or scatter graphs.

You might also not see any chart if the widgets were created a long time ago and they just fail to render (see the console in that case).

I just checked again on my non-retina laptop and the charts are correctly displayed in both views. Can you review on your side if the chart of the second, third and fourth widgets are not showing because they are overflowing?

pablopareja:
You're right, a couple of them are scrollable and the other one is failing for some reason...
Let's merge this and see if it keeps happening with new widgets 😉

PR #293, opened 2017-10-04
Fix/widget creation

mbarrenechea:
This PR adds: 
- Better validations for widgetEditor
- Legend buttons `type="button"`
- Legend Max width


mbarrenechea:
Yes, it's a problem with the layer. Did you try to save it????

pablopareja:
yeah, I tried to save it, why?

PR #322, opened 2017-10-09
Editor: sync the title of the chart with the widget's name

clementprdhomme:
This PR makes sure the name of the widget is shown in the editor and that whenever it is updated, the title of the chart also is (the other direction is also true).

image

[Pivotal task](https://www.pivotaltracker.com/story/show/151751570)

### DIFF
--- a/components/admin/widget/form/steps/Step2.js
+++ b/components/admin/widget/form/steps/Step2.js
@@ -8,11 +8,15 @@ import Textarea from 'components/form/TextArea';
 import Checkbox from 'components/form/Checkbox';
 import WidgetEditor from 'components/widgets/editor/WidgetEditor';
 
+// Redux
+import { connect } from 'react-redux';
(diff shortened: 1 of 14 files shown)


components/admin/widget/form/steps/Step2.js
```diff
@@ -146,7 +151,18 @@ class Step2 extends React.Component {
...
-export default Step2;
+const mapStateToProps = state => ({
+  widgetEditor: state.widgetEditor
+});
+
+const mapDispatchToProps = dispatch => ({
+  setTitle: title => dispatch(setTitle(title))
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(Step2);
```
mbarrenechea: I'm afraid that we are not using anymore the `admin/widget` folder. So....this change is unnecessary

clementprdhomme:
@mbarrenechea I've just reversed the changes to the file.

PR #329, opened 2017-10-10
Prevent the content of the raster editor from overflowing

clementprdhomme:
This PR updates the layout of the raster editor so more content can fit and it doesn't overflow.

Side-by-side comparison of the editor before and after the changes

### DIFF
--- a/components/widgets/editor/raster/RasterChartEditor.js
+++ b/components/widgets/editor/raster/RasterChartEditor.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Autobind } from 'es-decorators';
 import { toastr } from 'react-redux-toastr';
 import isEmpty from 'lodash/isEmpty';
+import truncate from 'lodash/truncate';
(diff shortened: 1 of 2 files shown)


components/widgets/editor/raster/RasterChartEditor.js
```diff
@@ -143,24 +159,37 @@ class RasterChartEditor extends React.Component {
     const { loading, bands, error, bandStatsInfo, bandStatsInfoLoading } = this.state;
     const { band, mode, showSaveButton, hasGeoInfo } = this.props;
 
+    let description = band && band.description;
+    const longDescription = description && description.length > 250;
```
mbarrenechea: I think that it's better to use https://lodash.com/docs/4.17.4#truncate. As long as it will truncate test by words => see this example https://stackoverflow.com/a/38150679/3603884

clementprdhomme:
@mbarrenechea Just made the change. Didn't know about lodash' function, thanks.

PR #352, opened 2017-10-11
Widgets: embed widget created

mbarrenechea:
This PR is important because, before it, you can't update map widgets....

- [ ] embed widget embed (I know...)

### DIFF
--- a/components/admin/widgets/form/WidgetsForm.js
+++ b/components/admin/widgets/form/WidgetsForm.js
@@ -253,7 +253,7 @@ class WidgetsForm extends React.Component {
 
   onChange(obj) {
     const form = Object.assign({}, this.state.form, obj);
-    this.setState({ form }, () => { console.log(form); });
+    this.setState({ form });
   }
 
   onStepChange(step) {
--- a/components/admin/widgets/form/WidgetsForm.js
+++ b/components/admin/widgets/form/WidgetsForm.js
@@ -289,7 +289,7 @@ class WidgetsForm extends React.Component {
       case 'map':
         return !!layer;
       case 'embed':
-        return !!embed.valid;
+        return !!embed.src;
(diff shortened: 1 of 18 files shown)


clementprdhomme:
Please do not merge, this is a WIP.

clementprdhomme:
@davidsingal @pablopareja 

The PR is ready to be reviewed.

[Pivotal task 1](https://www.pivotaltracker.com/story/show/151480551)
[Pivotal task 2](https://www.pivotaltracker.com/story/show/151322902)

pablopareja:
I've just tested the "embed widget" page that you mentioned and the loader keeps spinning forever 
image

clementprdhomme:
I managed to replicate the issue. For some reasons, it doesn't appear on Safari.

I copied the code of the `EmbedWidget` without noticing we have two loading variables: one for the data and another for the rendering of the Vega chart. In the embed widgets, we don't have the latter so I just removed it.

I also made further changes with the favourite feature:
* I moved the code of the favourite feature to the store as actions (it also factors it)
* I removed the spinner when the user clicks the star because we instantaneously update the icon to the desired state and eventually revert it if it fails (optimistic UI)
* I removed the star if the user is not logged in, because no further action can be taken
* I added the favourite feature to the map embeds (it was missing)

In case the query to (de-)favourite the widget fails, the star icon isn't reverted so we don't need anymore to display a toastr, that's why I've also removed it.

[Pivotal task 3](https://www.pivotaltracker.com/story/show/152273523)

pablopareja:
@clementprdhomme Could you please confirm that this is the expected behavior? 

> I removed the star if the user is not logged in, because no further action can be taken

I sort of remember someone saying that we should keep it and redirect to the register page when the user is not logged in.

PR #361, opened 2017-10-13
Improve the position of the tooltip on the sides of the screen

clementprdhomme:
This PR improves the position of the tooltip when displayed on either side of the screen: the tooltip is not cut anymore but gets "pinned" to the side and its tip is moved proportionally to point the desired position.

This behaviour hasn't been applied to the top and the bottom of the screen because the tooltip opens in these directions. If needed, we could reverse its direction but I don't think it's currently necessary.

<p align="center">
screen shot 2017-10-13 at 11 05 25
</p>

[Pivotal task](https://www.pivotaltracker.com/story/show/151915521)

### DIFF
--- a/components/ui/Tooltip.js
+++ b/components/ui/Tooltip.js
@@ -10,6 +10,12 @@ class Tooltip extends React.Component {
   constructor(props) {
     super(props);
 
+    this.state = {
+      // Horizontal offset of the tooltip's tip from its
+      // initial position (center)
+      tipOffset: 0
+    };
+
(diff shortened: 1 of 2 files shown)


mbarrenechea:
There are datasets that are not showing tags
image

clementprdhomme:
@mbarrenechea The issue you're mentioning is not directly related to the tooltip. The component doesn't manage what's inside.

I can fix the issues of `DatasetTagsTooltip` in `develop` if you want.

PR #382, opened 2017-10-18
Fix eslint and prop validation errors in ExploreDetails

clementprdhomme:
This PR fixes two eslint and two prop validation errors as well as improves the errors shown to the user.

In addition, it simplifies two functions that either were related to one of the errors or that displayed non-descriptive errors to the user.

### DIFF
--- a/pages/app/ExploreDetail.js
+++ b/pages/app/ExploreDetail.js
@@ -646,8 +646,7 @@ const mapDispatchToProps = dispatch => ({
     new LayersService()
       .fetchData({ id: layerId })
       .then(layer => dispatch(setLayer(layer)))
-      // TODO: better handling of the error
-      .catch(err => toastr.error('Error', err));
+      .catch(() => toastr.error('Error', 'Unable to load the layer of the widget.'));
   },
   setTitle: title => dispatch(setTitle(title)),
   setTopicsTree: tree => dispatch(setTopicsTree(tree))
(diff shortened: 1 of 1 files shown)


pages/app/ExploreDetail.js
```diff
@@ -155,35 +154,17 @@ class ExploreDetail extends Page {
...
-          });
-        }
+      .then(res => res.map(val => val.dataset).slice(0, 7))
+      .then((ids) => {
+        if (ids.length === 0) return [];
+        return DatasetService.getDatasets(ids, 'widget,metadata,layer,vocabulary');
       })
-      .catch(err => toastr.error('Error', err));
+      .then(similarDatasets => this.setState({ similarDatasets }))
+      .catch(() => toastr.error('Error', 'Unable to load the similar datasets'))
```
pablopareja: Could you also log the err object to the error console here?

pages/app/ExploreDetail.js
```diff
@@ -670,8 +646,7 @@ const mapDispatchToProps = dispatch => ({
     new LayersService()
       .fetchData({ id: layerId })
       .then(layer => dispatch(setLayer(layer)))
-      // TODO: better handling of the error
-      .catch(err => toastr.error('Error', err));
+      .catch(() => toastr.error('Error', 'Unable to load the layer of the widget.'));
```
pablopareja: same thing as what I mentioned above

pages/app/ExploreDetail.js
clementprdhomme: Yes! Actually that was one of my doubts. Don't you think a `console.error` is better than a `console.log`?

PR #387, opened 2017-10-20
Bypass the basic auth for the headless chrome UA

clementprdhomme:


davidsingal:
This answer is correct! The prize is for you!

PR #409, opened 2017-11-07
Let the user sort the datasets in Explore

clementprdhomme:
This PR lets the user sort the datasets in Explore by last modified date (default), most viewed or most favourited. Anytime the user changes the sorting order, the URL is updated so it can be restored later.

In addition, every time a user visits the Explore details page, a query is made to count a view to the dataset.

Filters of the datasets with a selector for the sorting order

While testing, I noted that the endpoints used to get the order of the datasets are returning a small number of datasets which aren't displayed in Explore, so changing the sorting order won't actually change the cards order.

@pablopareja Any idea why we don't get useful answers?

[Pivotal task 1](https://www.pivotaltracker.com/story/show/152429987)
[Pivotal task 2](https://www.pivotaltracker.com/story/show/152429883)

### DIFF
--- a/components/app/explore/DatasetListHeader.js
+++ b/components/app/explore/DatasetListHeader.js
@@ -1,10 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+
+// Components
 import Button from 'components/ui/Button';
 import Icon from 'components/ui/Icon';
-import { connect } from 'react-redux';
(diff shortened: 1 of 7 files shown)


clementprdhomme:
Hey @pablopareja, stupid mistakes. Everything seems to work fine now.

Thank you for the review.

pablopareja:
Cool, it seems to be working fine now 👏

PR #413, opened 2017-11-09
Use Transifex' locale in the some endpoints

clementprdhomme:
This PR saves Transifex' locale in the store (under `store.common.locale`) and use it in several API calls: the ones related to the datasets and the creation/update of subscriptions.

The metadata edition don't use the locale because it has a separate selector in the form. Generally speaking, the admin section doesn't use Transifex so the locale is always `en`.

Two comments:
1. I could not test Transifex properly because 1) it doesn't launch on localhost and 2) the only locale we have set in Transifex is `en`. We need to test in staging and add another locale to see if it's correctly passed to the API.
2. When the user changes the locale, the store is updated but the page's queries that were made  with the previous locale are not re-executed. Should we reload the page whenever the locale is changed? That would be the easiest solution.

[Pivotal task](https://www.pivotaltracker.com/story/show/152069427)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,7 +7,8 @@
   },
   "globals": {
     "L": true,
-    "config": true
+    "config": true,
+    "Transifex": true
   },
   "extends": "vizzuality",
   "rules": {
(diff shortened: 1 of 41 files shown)


pablopareja:
@clementprdhomme Yeah I agree that reloading the page would be a good solution. Could you add those changes before I carry out the merge?

clementprdhomme:
Once it's merged, do you think we can add a new language in Transifex to test if everything's working properly? The new language would also appear in staging.

PR #419, opened 2017-11-14
Fix undefined ENV variables in browser

clementprdhomme:
This PR fixes a bug where the newest ENV variables `API_ENV` and `GOOGLE_ANALYTICS` would be `undefined` in the browser causing what seems to be unrelated issues.

[Pivotal task (of a bug caused by this)](https://www.pivotaltracker.com/n/projects/1374154/stories/152806171)


clementprdhomme:
The variable `GOOGLE_ANALYTICS` is for the PR #415, but as this one might be merged before the other one, I've included the variable already.

pablopareja:
@clementprdhomme  Could you please also add the vars to the docker files?
Thanks!

PR #427, opened 2017-11-16
Feature/dashboards

mbarrenechea:
## Overview
New Wysiwyg added to dashboard form.

## Testing instructions
Remeber to run `yarn install`.

image


pablopareja:
There are a couple of problems related to styles in the widget selector.

The widgets overlap in some cases:
image

When there's scroll, I think we should keep the header fixed at the top and just scroll the component that holds the widgets
image

What do you think?

pablopareja:
One more thing, showing all widgets at the same time is way too much, we should add pagination to this component

mbarrenechea:
I agree with most of your comments. But I think that we should deploy as it is now, then create several improve tasks in pivotal to handle all of these requests.

PR #444, opened 2017-11-28
My RW profile form

mbarrenechea:
## Overview
- Add avatar (if exists) to header
- Add picture to form
- Remove change password from form

## Pivotal task
https://www.pivotaltracker.com/story/show/152028481


pablopareja:
I uploaded a picture (in theory successfully) but then I can't see it anywhere even if I log out and log back in again... 🤔

pablopareja:
Could you also please hide the "Delete account" checkbox ?
It shouldn't be there, at least as long as it doesn't work

mbarrenechea:
After checking it I talked with Raul => it's related to the API. So I think it's ready to be merged. He will fix it as soon as posible

pablopareja:
ok 👍

PR #468, opened 2017-12-12
Explore detail: Open in Explore button

pablopareja:
image

## Testing instructions

Check any dataset detail page that has at least one layer.

[Pivotal task](https://www.pivotaltracker.com/story/show/153453387)

### DIFF
--- a/pages/app/ExploreDetail.js
+++ b/pages/app/ExploreDetail.js
@@ -343,7 +343,7 @@ class ExploreDetail extends Page {
 
   handleOpenInExplore() {
     const { dataset } = this.state;
-    this.props.toggleLayerGroup(dataset.id, true, false);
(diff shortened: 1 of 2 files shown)


pages/app/ExploreDetail.js
```diff
@@ -334,6 +341,11 @@ class ExploreDetail extends Page {
     this.props.setModalOptions(options);
   }
 
+  handleOpenInExplore() {
```
mbarrenechea: As long as is a link to another page, should we use a Link or a Router.push? 
You can pass params:
- `<Link route="explore" params={{ layers: [dataset.id] }}/>` or
- `Router.pushRoute('explore', { layers: [dataset.id] });`

What do you think? Do you think it could work? That way you will avoid ```replaceRoute = true```

PR #469, opened 2017-12-12
Explore detail fix: datasets without tags

pablopareja:
Prior to this fix an error was thrown when trying to load the inferred tags of the dataset when it had not tags associated to it yet.

## Testing instructions
Check the Explore detail page of any dataset that has no tags associated to it.

## Pivotal task
None, I found the bug by chance.

### DIFF
--- a/pages/app/ExploreDetail.js
+++ b/pages/app/ExploreDetail.js
@@ -165,6 +165,7 @@ class ExploreDetail extends Page {
       loading: true
     }, () => {
       this.datasetService.fetchData('layer,metadata,vocabulary,widget').then((response) => {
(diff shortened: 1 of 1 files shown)


pages/app/ExploreDetail.js
```diff
@@ -165,6 +165,7 @@ class ExploreDetail extends Page {
       loading: true
     }, () => {
       this.datasetService.fetchData('layer,metadata,vocabulary,widget').then((response) => {
+        console.log('response', response);
```
mbarrenechea: Remove ```console.log```

pablopareja:
done 😉

PR #471, opened 2017-12-12
Explore: Search improvements

pablopareja:
The explore search now looks for text in the following dataset fields:
- Metadata title _(or dataset name if the latter doesn't exist)_
- Metadata info source
- Metadata info functions
- Tags

### Testing instructions
Search by values present in the fields: `source`, `functions` or tag names in the Explore page.

[Pivotal task](https://www.pivotaltracker.com/story/show/153594389)


pablopareja:
@mbarrenechea the search allows the use of special characters now

PR #478, opened 2017-12-15
Back office new feature: Subscribable datasets

pablopareja:
This PR allows admin users to set a dataset as subscribable by creating an arbitrary set of queries that can then be used by users to subscribe to the dataset.
> At least by now this can only be done for CARTO datasets.

image

### Testing instructions
Check this CARTO TEST dataset: http://localhost:3000/admin/data/datasets/e66092f9-a6a9-4a08-a787-7920de3fc934

[Pivotal Task](https://www.pivotaltracker.com/story/show/153454492)


components/datasets/form/DatasetsForm.js
```diff
@@ -117,11 +117,16 @@ class DatasetsForm extends React.Component {
             omit: (dataset) ? ['connectorUrlHint', 'authorization', 'connectorType', 'provider'] : ['connectorUrlHint', 'authorization']
           };
 
+          const bodyObj = omit(this.state.form, requestOptions.omit);
+          const newSubscribable = {};
```
mbarrenechea:
```
          bodyObj.subscribable = bodyObj.subscribable.reduce((o, val) => ({
            ...o,
            [val.type]: val.value
          }), {});
```

mbarrenechea: I think that this is beautiful and without eslint issues...

css/components/admin/dataset/form.scss
```diff
@@ -0,0 +1,11 @@
+.c-datasets-form {
+
+  .subscribable-row {
```
mbarrenechea: This is wrong. Look for Metadata form and field Sources. You will see that you only need to set the button margin-top => https://github.com/resource-watch/resource-watch/blame/feature/subscribable-dataset/css/components/admin/dataset/source-item.scss#L1-L6

components/datasets/form/steps/Step1.js
```diff
@@ -57,6 +64,42 @@ class Step1 extends React.Component {
     this.props.onChange({ legend });
   }
 
+  onSubscribableChange(obj) {
+    const { subscribable } = this.props.form;
```
mbarrenechea:
```
    const subscribable = this.props.form.subscribable.slice(0);
    const index = subscribable.findIndex(s => s.id === obj.id);

    subscribable[index] = {
      ...subscribable[index],
      ...obj
    };

    this.props.onChange({ subscribable });

```

mbarrenechea: This function is throwing an error because you are using the id as an index. What happens if I create an element, then create another one but I decide to remove the first one. ERROR if I try to edit the second I created. Try to use the function above

components/datasets/form/steps/Step1.js
```diff
@@ -57,6 +64,42 @@ class Step1 extends React.Component {
...
+    const newSubscribable = subscribable.slice(0);
+    newSubscribable.splice(index, 1)
+    this.props.onChange({ subscribable: newSubscribable });
+  }
+
+  handleAddSubscription() {
+    const { subscribable } = this.props.form;
+    const newSubscribable = subscribable.slice(0);
+    newSubscribable.push({ type: '', value: '', id: subscribable.length })
```
mbarrenechea: What happens if I create an element, then create another one but I decide to remove the first one. Then try to create another one, the would have the same id... => ERROR. Use Date.now() to be sure that they will be different

components/datasets/form/steps/Step1.js
```diff
@@ -57,6 +64,42 @@ class Step1 extends React.Component {
...
+    });
+    if (checked) {
+      this.props.onChange({ subscribable: [{ type: '', value:'', id: 0 }] });
+    } else {
+      this.props.onChange({ subscribable: [] });
+    }
+  }
+
+  handleRemoveSubscription(index) {
+    const { subscribable } = this.props.form;
```
mbarrenechea: The same here. You are assuming that the index would be the id. This is not true...

mbarrenechea:
```
  handleRemoveSubscription(id) {
    const subscribable = this.props.form.subscribable.slice(0);
    const index = subscribable.findIndex(s => s.id === id);

    subscribable.splice(index, 1)
    this.props.onChange({ subscribable });
  }

```

components/datasets/form/steps/Step1.js
mbarrenechea:
```
  handleAddSubscription() {
    const { subscribable } = this.props.form.subscribable.slice(0);
    subscribable.push({ type: '', value: '', id: Date.now() });
    this.props.onChange({ subscribable });
  }

```

components/datasets/form/steps/Step1.js
```diff
@@ -107,6 +150,10 @@ class Step1 extends React.Component {
 
     const columnFieldsOptions = (columnFields || []).map(f => ({ label: f, value: f }));
 
+    const addSubscribableButtonClassname = classnames({
```
mbarrenechea: Remove this

components/datasets/form/DatasetsForm.js
pablopareja: 👍

components/datasets/form/steps/Step1.js
pablopareja: 👍

components/datasets/form/steps/Step1.js
pablopareja: 👍

components/datasets/form/DatasetsForm.js
pablopareja: 👍

components/datasets/form/steps/Step1.js
pablopareja: I've used a counter better than Date.now(), WDYT?
``` this.subscribableCounter``` that I'm initializing in the constructor and updating everytime the add button is clicked

Comment (mbarrenechea) [reply components/datasets/form/steps/Step1.js]: I don't like it. What would happen if I edit one subscribable dataset that already exists? Would the counter be initialized at 0? What if there is already a 0 id? You need to be sure that the id is new and unique. Time is unique and I think that's better than add a library or something else

**Comment by pablopareja** [comment]
@mbarrenechea could you have a quick look at the new styles for the subscribable fields?
I don't understand why the rows are taking more space than what they should...
image

File: components/datasets/form/steps/Step1.js [hunk 7]
```diff
@@ -85,10 +111,24 @@ class Step1 extends React.Component {
...
+    const subscribable = this.props.form.subscribable.slice(0);
+    const index = subscribable.findIndex(s => s.id === id);
+    subscribable.splice(index, 1);
+    this.props.onChange({ subscribable });
+  }
+
+  handleAddSubscription() {
+    const subscribable = this.props.form.subscribable.slice(0);
+    subscribable.push({ type: '', value: '', id: Date.now() });
+    this.subscribableCounter++;
```
mbarrenechea: Remove this

mbarrenechea:
The reason is simple. col-3 + col-7 = 10 but then you put a button without col.... result = 10 + button width

PR #483, opened 2017-12-20
Back office: Widget metadata + widget links

pablopareja:
This PR has added a metadata section for widgets in the back office. At the moment the metadata section only includes the `widgetLinks` field.

### Testing instructions
The following test widget can be used: `http://localhost:3000/admin/data/widgets/b997f049-3a8a-4648-8b63-192f26800eda/metadata`

[Pivotal task](https://www.pivotaltracker.com/story/show/153454404)

image


components/admin/widgets/metadata/form/MetadataForm.js
```diff
@@ -0,0 +1,186 @@
+import React from 'react';
```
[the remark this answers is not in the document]
mbarrenechea: Also, I don't see anywhere the new structure that we talked about. We talked that we were going to use Victor's way to add new components. What I mean is that you still make calls from the component instead of using actions. I know that change forms now would be complicated but please, for new components try to use it. Ok?

components/admin/widgets/metadata/form/constants.js
```diff
@@ -0,0 +1,60 @@
...
+    const valid = Object.keys(elements)
+      .map(k => elements[k].isValid())
+      .filter(v => v !== null)
+      .every(element => element);
+
+    return valid;
+  }
+};
+
+export const SOURCE_ELEMENTS = {
```
mbarrenechea: Are you using this?

css/components/admin/widget/metadata_form.scss
```diff
@@ -0,0 +1,18 @@
+.c-widget-metadata-form {
+  
+  .remove-widget-link-container {
```
mbarrenechea: We would avoid this kind of style by using a component, but it's your call....

services/WidgetsService.js
```diff
@@ -52,10 +52,10 @@ export default class WidgetsService {
     });
   }
 
-  fetchData({ id }) {
+  fetchData({ id, includes = '' }) {
     return new Promise((resolve, reject) => {
       get({
-        url: `${process.env.WRI_API_URL}/widget/${id}`,
+        url: `${process.env.WRI_API_URL}/widget/${id}?includes=${includes}`,
```
mbarrenechea: We were talking about this. I think that it would be better if you handle with 'query-string'. There is an example here => https://github.com/resource-watch/resource-watch/blob/develop/components/dashboards/wysiwyg/widget-block-edition/widget-block-edition-actions.js#L21

components/admin/widgets/metadata/form/MetadataForm.js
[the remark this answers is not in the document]
pablopareja: Ok, I've just done the refactor of the metadata components in this commit: bbbc8255e384ab83a472d540f94844525426a1a1 
About your second comment, ok that makes sense 👍

components/admin/widgets/metadata/form/constants.js
pablopareja: Not at the moment but it will be used as the number of fields here grows...

services/WidgetsService.js
pablopareja: cool, thanks for mentioning that. I'll try to use from now on.
It's been included in this commit: a9aa17edc162db40a82405361b32d5b13815f987

components/admin/widgets/metadata/form/steps/Step1.js
```diff
@@ -0,0 +1,191 @@
...
+    * UI EVENTS
+    * - onWidgetLinkChange
+    * - onWidgetLinksCheckboxChange
+    * - handleRemoveWidgetLink
+    * - handleAddWidgetLink
+  */
+  onWidgetLinkChange(obj) {
+    const widgetLinks = this.props.form.info.widgetLinks.slice(0);
+    const index = widgetLinks.findIndex(elem => elem.id === obj.id);
+    widgetLinks[index] = {
```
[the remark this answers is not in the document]
pablopareja: yeah I know... let's see if I can find the time to create it soon enough 😉

css/components/admin/widget/metadata_form.scss
pablopareja: 👍

PR #484, opened 2017-12-20
Feature/explore visual updates

pablopareja:


pablopareja:
@mbarrenechea could you have a look at how to fix the grid?
I've been doing some tests with no success trying to get foundation to use breakpoints values different from the default ones...

pages/app/Explore.js
```diff
@@ -516,14 +551,10 @@ class Explore extends Page {
...
     const { topics, geographies, dataType } = this.filters;
-    const topicsLabels = topics.map(topic => findTagInSelectorTree(topicsTree, topic).label);
-    const geographiesLabels = geographies.map(geography =>
-      findTagInSelectorTree(geographiesTree, geography).label);
-    const dataTypeLabels = dataType.map(dType => findTagInSelectorTree(dataTypeTree, dType).label);
 
-    const allTagsSt = [].concat(topicsLabels).concat(geographiesLabels)
-      .concat(dataTypeLabels).join(', ');
-    const filtersSumUp = !showFilters && allTagsSt.length > 0 ? `Filtering by ${allTagsSt}` : '';
+    const selectedTags = [...topics.map(t => ({ ...findTagInSelectorTree(topicsTree, t), type: 'topics' })),
```
mbarrenechea: I think this is the perfect example to create a component. Don't you feel overwhelmed when you enter Explore page and see 878 lines of code? We should avoid loading everything on the page. That's why we need to create components.

Here you have an example => https://github.com/resource-watch/resource-watch/tree/develop/components/dashboards/wysiwyg/widget-block-edition

pages/app/Explore.js
```diff
@@ -439,6 +441,39 @@ class Explore extends Page {
     }
   }
 
+  handleRemoveTag(tag) {
```
mbarrenechea: These functions would be better in a Component...

css/components/app/pages/explore.scss
```diff
@@ -31,7 +31,41 @@
   }
 
   .filters-sum-up {
-    color: $dark-pink;
+    display: flex;
```
mbarrenechea: the same as the other comments. It should be a component, explore.scss is not a file where you can add whatever you want

components/app/explore/explore-dataset-filters/explore-dataset-filters-actions.js
```diff
@@ -0,0 +1,21 @@
+import { createAction, createThunkAction } from 'redux-actions';
+
+import FiltersService from 'services/FiltersService';
+
+export const setDataFilters = createAction('explore-dataset-filters/setDataFilters');
+export const setFilter = createAction('explore-dataset-filters/setFilter');
```
mbarrenechea: Loading an Error actions `setFilterLoading` and `setFilterError`

mbarrenechea: I miss the default state. I recommend you to start always by defining the default state because it will give a good shape of what you are going to achieve.

mbarrenechea: Remember that you can't add logic inside the `xxx-component.js`

components/app/explore/explore-dataset-filters/explore-dataset-filters.js
```diff
@@ -0,0 +1,33 @@
...
+
+import ExploreDatasetFiltersComponent from './explore-dataset-filters-component';
+import getFilterStatus from './explore-dataset-filters-selectors';
+
+const mapStateToProps = state => ({
+  data: getFilterStatus(state)
+});
+
+class ExploreDatasetFiltersContainer extends Component {
+  componentWillMount() {
```
davidsingal: You are requesting filters here, I think you have to use componenDidMount. Are you using server rendering to get filters? In that case, you don't need it.

PR #485, opened 2017-12-21
Dataset Tags form fixes

pablopareja:
This PR fixes the Dataset tags form _(prior to this clicking in the button "Save" had no effect whatsoever)_
The request used to update the tags of a dataset has also been updated so that it uses *DELETE* when it's sending an empty array of tags.


pablopareja:
@mbarrenechea Ideally I guess that we should create a "Step1" component for this, but this fix get things back to work in much less time. _(We could put that improvement in the backlog)_

PR #487, opened 2017-12-27
Back office: load only datasets that were created by users with an ADMIN role

pablopareja:
### Testing instructions
Check the datasets table `http://localhost:3000/admin`

[Pivotal task](https://www.pivotaltracker.com/story/show/153542866)

### DIFF
--- a/services/DatasetsService.js
+++ b/services/DatasetsService.js
@@ -19,11 +19,12 @@ export default class DatasetsService {
       application: applications.join(','),
       language: this.opts.language,
       ...!!includes && { includes },
-      'user.role': 'ADMIN',
       'page[size]': 9999999,
       env: process.env.API_ENV,
       ...filters
     };
+
+    console.log(qParams, filters);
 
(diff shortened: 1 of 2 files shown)


services/DatasetsService.js
```diff
@@ -19,6 +19,7 @@ export default class DatasetsService {
       application: applications.join(','),
       language: this.opts.language,
       ...!!includes && { includes },
+      'user.role': 'ADMIN',
```
davidsingal: For now, it is used only in the Backoffice but we should pass this as param from the component.

pablopareja: 👍

services/DatasetsService.js
```diff
@@ -24,6 +24,8 @@ export default class DatasetsService {
       ...filters
     };
 
+    console.log(qParams, filters);
```
davidsingal: Remove it :)

pablopareja: done 😉

PR #490, opened 2017-12-28
Fix for Explore Detail page

pablopareja:
## Overview
Fix for similar datasets tags tooltip _(some obsolete code has also been removed)_

## Testing instructions
Check the Explore detail page of any dataset, for instance: `http://localhost:3000/data/explore/63f34231-7369-4622-81f1-28a144d17835`. Go to the similar datasets section, click on the tags icon and click in any of the tag pills shown.

@davidsingal  There's a problem with the load of the Explore page though... it loads with a huge scroll when, in theory the page should not have any scroll at all 🤔 
It looks like it somehow _inherits_ the page size of the Explore Detail page. 

## Pivotal task
None, bug found by @pablopareja

### DIFF
--- a/pages/app/ExploreDetail.js
+++ b/pages/app/ExploreDetail.js
@@ -8,7 +8,7 @@ import classnames from 'classnames';
 // Redux
 import withRedux from 'next-redux-wrapper';
 import { initStore } from 'store';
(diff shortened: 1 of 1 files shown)


davidsingal:
Normally It happens when you are navigation from a large page. Does it?

PR #498, opened 2018-01-12
Surface widget links

pablopareja:
## Overview
This PR surfaces to app site the widget links that can be created from the metadata section in the back office. 
They are visible now in the following two pages:

1. `Profile --> My Widgets` section as part of the Widget Actions tooltip:

image

2. Embed Widget page in the flipcard

image

> NOTE: The title hyperlink of the widget embed page is deactivated when there exist `widgetLinks` defined.

## Testing instructions

Check the two pages mentioned above using a widget for which you've defined a set of widget links.

### DIFF
--- a/components/widgets/list/WidgetActionsTooltip.js
+++ b/components/widgets/list/WidgetActionsTooltip.js
@@ -48,6 +48,8 @@ class WidgetActionsTooltip extends React.Component {
   }
 
   render() {
+    const { widgetLinks } = this.props;
+    console.log('widgetLinks', widgetLinks);
     return (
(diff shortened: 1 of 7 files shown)


components/widgets/list/WidgetActionsTooltip.js
```diff
@@ -48,6 +48,8 @@ class WidgetActionsTooltip extends React.Component {
   }
 
   render() {
+    const { widgetLinks } = this.props;
+    console.log('widgetLinks', widgetLinks);
```
mbarrenechea: 👎

PR #504, opened 2018-01-15
Improve the tooltip of the Vega Chart

clementprdhomme:
## Overview
This PR contains changes merged in PREP. See the details here: https://github.com/resource-watch/prep-manager/pull/31

## Testing instructions
There are currently not test cases in RW.

## Pivotal task
None.

### DIFF
--- a/components/widgets/charts/VegaChartTooltip.js
+++ b/components/widgets/charts/VegaChartTooltip.js
@@ -25,41 +25,43 @@ class VegaChartTooltip extends React.Component {
     return x.value;
   }
 
-  getParsedY() {
-    const { item } = this.props;
-    if (!item) return null;
-
-    const { y } = item;
-    if (!y) return null;
-
-    if (y.format && y.type === 'number') {
-      return format(y.format)(y.value);
-    } else if (y.type === 'date') {
-      const date = new Date(y.value);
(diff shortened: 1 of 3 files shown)


components/widgets/charts/VegaChartTooltip.js
```diff
@@ -25,60 +25,59 @@ class VegaChartTooltip extends React.Component {
...
   }
 }
 
-const columnType = PropTypes.shape({
-  type: PropTypes.oneOf(['number', 'string', 'date']),
-  label: PropTypes.string,
-  format: PropTypes.string,
-  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.object])
-});
+// const columnType = PropTypes.shape({
```
mbarrenechea: If we are not using it, remove it. If you want to preserve it for future implementations don't read this....oh!! you already did it...

pablopareja:
@clementprdhomme The tooltip seems to appear displaced

components/widgets/charts/VegaChartTooltip.js
clementprdhomme: Actually, I wanted to still use it but `prop-types` don't let me define the props as I wanted. I just let the text there so we remember the structure in the future.

PR #506, opened 2018-01-16
Timelines: layer order

pablopareja:
## Overview
This PR adds the possibility to display in Explore the timeline of any given dataset having a set of layers with the following attributes inside the `layerConfig` object:

- `timeline`: should be equals to `true` for the layers that should be part of the timeline
- `order`: number that will be used to calculate the position of the layer in the timeline

## Testing instructions
Check any of these two datasets in Explore:
- c0c71e67-0088-4d69-b375-85297f79ee75
- b8307c16-fd77-4e35-9b68-8726a025f401
or experiment with other datasets adding the _*timeline*_ capability to them following the instructions mentioned above.


nsuberi:
Hello, it seems that you also must have a dateTime property in the LayerConfig object in order to work correctly.

PR #510, opened 2018-01-16
Improve the tooltip of the Vega Chart

clementprdhomme:
## Overview
This PR contains changes merged in PREP. See the details here: https://github.com/resource-watch/prep-manager/pull/31

**Previous PR #504 was merged by error and then reverted.**

## Testing instructions
There are currently not test cases in RW.

## Pivotal task
None.

### DIFF
--- a/components/widgets/charts/VegaChart.js
+++ b/components/widgets/charts/VegaChart.js
@@ -144,9 +144,11 @@ class VegaChart extends React.Component {
 
     // If the cursor is on top of a mark, we display the data
     // associated to that mark
-    // The only exception is for the lines because the data
-    // they own is always the first one (first point)
-    if (!isEmpty(item) && item.datum.x && item.mark.marktype !== 'line') {
+    // The only exception is for the lines, areas and text because the
+    // data they own is always the first one (first point)
+    if (!isEmpty(item) && item.datum.x && item.mark.marktype !== 'line'
+      && item.mark.marktype !== 'area'
+      && item.mark.marktype !== 'text') {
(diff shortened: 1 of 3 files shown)


clementprdhomme:
@pablopareja Regarding your comment on the previous PR, can you give me a URL where I can reproduce the issue?

The changes have no impact on the position of the tooltip, only its content. I've just reviewed again and I don't see any issue.

pablopareja:
@clementprdhomme check this widget for instance: http://localhost:3000/embed/widget/b997f049-3a8a-4648-8b63-192f26800eda
Perhaps the problem is something that was already happening before this change but we should fix it anyway...

pablopareja:
image

pablopareja:
image

clementprdhomme:
Big fan of the photos 😂.

First and foremost, the issue you're seeing does not come from this PR so we should deal with it somewhere else.

Unfortunately, the problem cannot be solved as it comes from Vega. Under the hood, the width of the bars is auto calculated by Vega and then the bars are horizontally displaced by 15px to the left so they are centred with the ticks. It seems that for the `mousemove`event, Vega actually sees the bars as if the offset didn't exist. The chances this bug gets fixed are really slim because v2 of Vega hasn't been updated in 6 months.

The good news is that if we go with our plan to migrate to v3, the bug will be probably be fixed (🤞).

pablopareja:
hehehe well it was the only way I could think of to show where the cursor was... 😂 
Ok I understand, let's move on with this PR and see if we can adopt Vega 3 soon enough

PR #524, opened 2018-01-24
Embed similar datasets page

pablopareja:
## Overview
This PR adds the component *'similar datasets'* that is used from now on both in the new EmbedSimilarDatasets page and in Explore detail.

## Testing instructions
Check the Embed page for any given dataset, e.g. http://localhost:3000/embed/similar_datasets/fa6443ff-eb95-4d0f-84d2-f0c91682efdf

## [Pivotal task](https://www.pivotaltracker.com/story/show/154564392)

image


components/app/explore/similar-datasets/similar-datasets-component.js
```diff
@@ -0,0 +1,46 @@
...
+    const { similarDatasets } = this.props;
+    const { loading, data, error } = similarDatasets;
+
+    return (
+      <div className="c-similar-datasets">
+        <Spinner isLoading={loading} className="-light" />
+        <h3>Similar datasets</h3>
+        {data && data.length > 0 &&
+          <DatasetList
+            active={[]}
```
mbarrenechea: Should `DatasetList`have a default prop active set as an empty array instead of passing it empty?

pablopareja: yeah, I agree 👍

components/app/explore/similar-datasets/similar-datasets-actions.js
```diff
@@ -0,0 +1,27 @@
...
+export const getSimilarDatasetsError = createAction('similar-datasets/getSimilarDatasetsError');
+export const setSimilarDatasets = createAction('similar-datasets/getSimilarDatasets');
+
+// Async actions
+export const getSimilarDatasets = createThunkAction('similar-datasets/getSimilarDatasets', (datasetId, locale = 'en') => (dispatch) => {
+  dispatch(getSimilarDatasetsLoading());
+  const service = new DatasetService(datasetId, { apiURL: process.env.WRI_API_URL, language: 'en' });
+  return service.getSimilarDatasets()
+    .then((data) => {
+      DatasetService.getDatasets(data.map(d => d.dataset), locale, 'widget,metadata,layer,vocabulary')
```
mbarrenechea: Only one more question: Why are we using `service`but after that we use an static method `DatasetService.getDatasets`?

PR #537, opened 2018-01-29
Splash coral: all hotspots missing added plus latest design updates applied

pablopareja:
## Overview
This PR adds the text and markup for the new three hotspots added to the _healthy_ layer of the coral splash.

## Testing instructions
Check http://localhost:3000/splash/coral and select the _healthy_ layer.

## [Pivotal task](https://www.pivotaltracker.com/story/show/154715436)

image


utils/splash/Panoramas.js
```diff
@@ -10,22 +10,77 @@ export const PANORAMAS = [
...
+                <p>of ocean species <a href="http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0025026" target="_blank" rel="noopener noreferrer">depend on tropical reefs.</a></p>
+                <h3>$29.8 billion</h3>
+                <p>is the <a href="http://www.icriforum.org/sites/default/files/Economic_values_global compilation.pdf" target="_blank" rel="noopener noreferrer">estimated yearly value</a> of all the services reefs provide, including tourism, coastal protection, fisheries and biodiversity</p>
+                <h3>500 million</h3>
+                <p>people <a href="http://www.noaa.gov/explainers/coral-reefs-essential-and-threatened" target="_blank" rel="noopener noreferrer">depend on coral reefs</a> for their food and livelihoods</p>
+                <h3>75</h3>
+                <p>percent of reefs worldwide <a href="http://www.reefresilience.org/pdf/Reefs_at_Risk_Revisited.pdf" target="_blank" rel="noopener noreferrer">are threatened</a> by warmer oceans and local threats such as coastal development, overfishing and pollution</p>
+              </div>,
+            image: '../../static/images/splash/coral_bleaching_on_the_rise.png',
```
mbarrenechea: Do you really need to set this path as relative? I'm just asking if you should use this one `/static/images/splash/coral_bleaching_on_the_rise.png`

pablopareja: I've just made a quick test and it seems to work as well if we use `/static/...` so I will update them.
I think we were using this relative path because we had problems in the past when referencing absolute ones with Aframe _(at least inside Aframe entities)_

PR #552, opened 2018-02-02
Partner detail: Add related datasets

pablopareja:
## Overview
This PR adds a related content section - datasets - to the partner detail pages.

## Testing instructions
Check this link for instance: http://localhost:3000/about/partners/wri

## [Pivotal task](https://www.pivotaltracker.com/story/show/153542953)

image


pages/app/PartnerDetail.js
```diff
@@ -15,6 +27,11 @@ class PartnerDetail extends Page {
   */
   componentDidMount() {
     this.props.getPartnerData(this.props.url.query.id);
```
mbarrenechea: I think we should use server rendering here => `this.props.getPartnerData(this.props.url.query.id)`

PR #590, opened 2018-02-21
Feature/layer interactivity

tomasmoose:
## Overview
Add a simple ui so we can edit the layer interactivity

## Testing instructions
This url is good for testing: http://localhost:9000/admin/data/layers/17b9bf19-e116-4a22-b71a-fe67ce7fd552/edit 

Remember to test layers that currently dont have a config

## Pivotal task
https://www.pivotaltracker.com/story/show/155226677


mbarrenechea:
I know that here we are not using redux, but as we want to share this component with other apps (PREP, future apps), at least be able to copy and paste it, we should try another approach. Let's talk tomorrow.

BTW, really good work

tomasmoose:
@mbarrenechea for monday

Its now using redux, 

For the format select input i added these options, we can add more if its needed: 

```
const DATA_FORMATS = {
  number: [
    { label: '00000', value: '00000' },
    { label: '0,0', value: '0,0' },
    { label: '0a', value: '0a' }
  ],
  date: [
    { label: 'YYYY/MM/DD', value: 'YYYY/MM/DD' },
    { label: 'YYYY', value: 'YYYY' },
    { label: 'MM', value: 'MM' },
    { label: 'DD', value: 'DD' }
  ]
};
```

tomasmoose:
@mbarrenechea now we are removing the references dynamically for the interactions!

mbarrenechea:
For the future =>
- Add drag and drop sorting. Check faqs

PR #593, opened 2018-02-22
add mainDateField for dataset form

tomasmoose:
## Overview
Add new field so its possible to edit mainDate in the dataset form

## Pivotal task
https://www.pivotaltracker.com/story/show/155358235


tomasmoose:
thanks ! will think of that :)

PR #598, opened 2018-02-23
Feature/dataset list

mbarrenechea:
## Overview
This PR just move dataset list component to `components/dataset`folder. 
It also add a custom grid behavior if you want. It will be very helpful for the responsive of this component.


css/layouts/_row.scss
```diff
@@ -0,0 +1,20 @@
+.l-row {
+  &.-equal-height {
+    margin-top: -#{map-get($grid-column-gutter, small)};
+
+    @media screen and (min-width: map-get($breakpoints, medium)) {
+      margin-top: -#{map-get($grid-column-gutter, medium)};
+    }
+
+    div[class*=col],
```
pablopareja: Quick question, `div[class*=col]` will select all divs that can be selected by `div[class^=col]`, so why adding both selectors?

PR #603, opened 2018-02-27
Join us button at footer and newsletter modal

pablopareja:
## Overview
This PR adds a button to the footer ("Join us") that opens a modal with the form to sign up to the newsletter.

## Testing instructions
Go to the footer of any page and click on "Join us"

### DIFF
--- a/components/modal/newsletter-modal/index.js
+++ b/components/modal/newsletter-modal/index.js
@@ -1,0 +1,11 @@
+import { connect } from 'react-redux';
+
+import NewsletterModalComponent from './component';
+
+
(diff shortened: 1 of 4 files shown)


components/modal/newsletter-modal/index.js
```diff
@@ -0,0 +1,11 @@
+import { connect } from 'react-redux';
+
+import NewsletterModalComponent from './component';
+
+
+export default connect(
+  state => ({
+    ...state.shareModal
```
mbarrenechea: You don't need redux in this component

PR #609, opened 2018-03-01
FIX/contact form submission and error handling

tomasmoose:
## Overview
Contact form had some issues, they will be resolved with this pr, also the button will get disabled when contacting, to avoid multiple requests

## Testing instructions
Fill the contact form with valid information, it should post > success > disable button 

start typing again in the form and the button should be enabled again

## Pivotal task
https://www.pivotaltracker.com/story/show/155362195

### DIFF
--- a/components/app/common/ContactUs/ContactUsForm.js
+++ b/components/app/common/ContactUs/ContactUsForm.js
@@ -47,7 +47,7 @@ class ContactUsForm extends React.Component {
           body: this.state.form
         })
           .then(() => {
-            this.setState({ submitting: false, submitted: true });
+            this.setState({ submitting: false });
             toastr.success('Success', 'Your message has been sent!');
           })
           .catch(() => {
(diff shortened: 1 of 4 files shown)


components/app/common/ContactUs/ContactUsForm.js
```diff
@@ -103,7 +103,7 @@ class ContactUsForm extends React.Component {
           </Field>
 
           <div className="actions-container -align-right">
-            <button type="submit" className="c-btn -primary">
+            <button type="submit" className={`c-btn -primary ${this.state.submitted ? '-disabled' : null}`} disabled={this.state.submitted}>
```
mbarrenechea: Should we disable the form after submitting or during the submit? I mean, I think that we should disable the button with the `this.state.submitting`

tomasmoose: makes sense, will change the logic so its disabled when submitting

PR #619, opened 2018-03-12
Fix Splash page --> Use Globe Cesium component

pablopareja:
## Overview
This PR fixes the issue related to using the old `Map` component in the Splash page.

## Testing instructions
Go to `localhost:3000/splash`

## [Pivotal task](https://www.pivotaltracker.com/story/show/155892238)


pages/app/Splash.js
```diff
@@ -17,14 +17,9 @@ import Header from 'components/splash/layout/Header';
 import { MARKERS } from 'utils/splash/Markers';
 
 // Components
+import GlobeCesium from 'pages/app/pulse/globe-cesium';
```
mbarrenechea: Are you getting a component from another page?? As you need to share it between pages you should move this component to the components folder.

pages/app/Splash.js
```diff
@@ -149,6 +155,7 @@ class Splash extends Page {
   }
 
   handleOnInit(viewer) {
+    console.log('onINit');
```
mbarrenechea: Remove it

PR #643, opened 2018-03-26
Widget Editor --> Title and caption inputted into the editor copy over to the save widget dialog

pablopareja:


components/modal/SaveWidgetModal.js
```diff
@@ -15,6 +15,7 @@ import Spinner from 'components/ui/Spinner';
 
 // Services
 import WidgetService from 'services/WidgetService';
+import WidgetsService from 'services/WidgetsService';
```
mbarrenechea: 2 services??

PR #666, opened 2018-04-03
Feature/pulse responsive

pjosh:
## Overview
These changes make the planet pulse responsive!
Now, we have a new native select that appears for small screen devices. A significant change that I've made is to move the toggle layers logic from the `layer-menu-dropdown` to the `layer-menu`, mainly to reuse the actions methods in the `layer-menu-native`.
Besides, I've added the `react-hammerjs` library to manage mobile gestures.

pulse

## Notes
We have a responsive issue with the choropleth legend. We get it from https://github.com/resource-watch/wri-api-components so I'm going to fix it there.
captura de pantalla 2018-04-03 a las 9 50 45


layout/pulse/layer-menu-native/component.js
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <div className="c-layer-menu-native">
+        <select
+          onChange={(event) => {
+            const { value } = event.target;
+            if (value === -1) {
+              triggerClick(null);
+            } else {
+              layers.forEach((g) => {
+                const index = findIndex(g.layers, layer => event.target.value === layer.label);
```
mbarrenechea: Do you know that we have a native findIndex => https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Objetos_globales/Array/findIndex 

It's true that we need a polyfill. But we are loading it https://github.com/resource-watch/resource-watch/blame/feature/pulse-responsive/layout/head/app.js#L158

layout/pulse/layer-menu-native/index.js
```diff
@@ -0,0 +1,24 @@
...
+import * as reducers from './reducer';
+import initialState from './initial-state';
+
+import LayerMenuNativeComponent from './component';
+
+const mapStateToProps = state => ({
+  layerActive: state.pulse.layerActive
+});
+
+class LayerMenuNativeContainer extends Component {
```
mbarrenechea: This is unnecessary, you can pass directly the component to the connect. In fact, you are using the same name...eslint should be saying something

pjosh:
done! Thanks @mbarrenechea !
I had never used `classnames` before, I love it! I want to use it in GFW from now on.

PR #676, opened 2018-04-03
Save & Remove from favorites 

tomasmoose:
## Overview
Dont use popup to save @ collection when clicking favorites icon, just save to favorites or remove directly. 

Bonus
+ generic linting stuff
+ Cleanup logic
+ use `response.ok` instead of checking for **200** status code for favorite requests

## Testing instructions
1. Make sure you can save datasets & widgets to favorites
2. Make sure you can remove them also

There should never appear any collection popup! 

## Pivotal task
https://www.pivotaltracker.com/story/show/156256396

### DIFF
--- a/components/favorites/ToggleFavorite.js
+++ b/components/favorites/ToggleFavorite.js
@@ -1,0 +1,100 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+// Redux
+import { connect } from 'react-redux';
+
+import { toggleFavourite } from 'redactions/user';
+
+// Selectors
+import { parseFavourites } from 'selectors/user/favorites';
+
+// helpers
+import { belongsToACollection } from 'components/collections-panel/collections-panel-helpers';
+
+// Components
+import Icon from 'components/ui/Icon';
+import LoginRequired from 'components/ui/login-required';
+
+class ToggleFavorite extends React.Component {
+  toggleFavorite() {
+    const {
+      toggleFavourite,
+      favourites,
+      data,
+      type,
+      user
+    } = this.props;
+
+    const favourite = favourites.find(fav => fav.resourceId === data.id) || {};
+
+    if (user.favourites.loading) return;
+
+    toggleFavourite({
+      favourite,
+      resource: {
+        resourceId: data.id,
+        resourceType: type
+      }
+    });
+  }
+
+  render() {
+    const {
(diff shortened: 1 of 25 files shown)


css/_base.scss
```diff
@@ -162,6 +162,15 @@ input, button, select, textarea {
   cursor: move;
 }
 
+.favourite-button {
```
mbarrenechea: I don't like this here in the `base.scss`.... Should we create a `c-favourite-button`? What do you think?

layout/explore/explore-datasets/explore-datasets-actions/index.js
```diff
@@ -8,5 +8,5 @@ export default connect(
   state => ({
     ...state.explore.map
   }),
-  actions
+  ...actions
```
mbarrenechea: Are you sure about this??

components/favorites/ToggleFavorite.js
```diff
@@ -0,0 +1,100 @@
...
+const mapStateToProps = state => ({
+  user: state.user,
+  favourites: parseFavourites(state)
+});
+
+const mapDispatchToProps = dispatch => ({
+  toggleFavourite: fav => dispatch(toggleFavourite(fav))
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(ToggleFavorite);
```
mbarrenechea: Split this component into a container and a component. Like we are doing with all of the components

layout/explore/explore-datasets/explore-datasets-actions/index.js
mbarrenechea: Are you sure about this?

css/_base.scss
mbarrenechea: Let's create a `c-favourite-button` instead of puting this here (_base.scss)... What do you think?

tomasmoose: makes sense, will move it out to its own file !

layout/explore/explore-datasets/explore-datasets-actions/index.js
tomasmoose: not so sure, fixing :D

css/_base.scss
tomasmoose: its a bit messy, we are shadow applying styles on all places where the favorite button is, so still needed to update lots of files, but well... its working a bit better now

PR #680, opened 2018-04-04
FEAT / custom layer interaction

tomasmoose:
## Overview
Allow admin users to add custom interactions, show warning so they know its a custom one when selecting format.

* Fix data type issues breaking the map tooltip
* When updating a layer, you now stay on the same page 
 * When entering layer admin, the tooltip now works when no interactions are present on init

Updated UI 

New button under layer config to save the form, also now we verify that the layer config is valid before submit.  (see error example in screenshot)
screen shot 2018-04-05 at 15 58 39

If you click "verify" layer config and its valid, you will get success feedback
screen shot 2018-04-06 at 12 07 54

Custom interactions now have a different UI as they can easily break the query. 
screen shot 2018-04-05 at 16 39 56

If you add a custom interaction, and try to save, while that field is NOT available in the **query** you wont be available to save the layer.

## Testing instructions
`/admin/data/layers` start typing in the field and if we dont have the interactions you should be prompted to add it. 

All formats should be available in the format select input for the custom ones, also when focusing you should see a warning that its a custom one without data type. 

## Pivotal task
https://www.pivotaltracker.com/story/show/156460988


mbarrenechea:
- Show when the layer is failing
- Custom interactions should show a warning
- If you come with no interactions tooltip is not working
- Refresh button is not working

components/admin/layers/form/LayersForm.js
```diff
@@ -116,39 +124,34 @@ class LayersForm extends React.Component {
...
-            .then((data) => {
-              toastr.success('Success', `The layer "${data.id}" - "${data.name}" has been uploaded correctly`);
-              if (this.props.onSubmit) this.props.onSubmit();
-            })
-            .catch((errors) => {
+        // Start the submitting
+        this.setState({ submitting: true });
+        this.props.dispatch(setLayerInteractionError(false));
+
+        if (cartoLayer.length) {
```
mbarrenechea: Should we think in the future? I mean, right now we only have cartodb interactions, but in the future we will add all kinds of layers. Should we have a utils file to verify every kind of layer? Don't do this right now, but next time it would be great

PR #683, opened 2018-04-04
FIX / issues with partner pages not showing their corresponding datasets

tomasmoose:
## Overview
- implement **WRISerializer** for partnerDetails, so we have the correct data format for the datasets.
- Add condition checks for widget, layer so the page does not break if `undefined` values are returned, this can happen on load
- Tall partner images overflowed their containers, added max height to prevent that
- Some linting while at it

## Testing instructions
`http://localhost:9000/about/partners/unep` for example, make sure you see the corresponding datasets

## Pivotal task
https://www.pivotaltracker.com/story/show/156482929

### DIFF
--- a/components/datasets/list/list-item/component.js
+++ b/components/datasets/list/list-item/component.js
@@ -27,7 +27,7 @@ class DatasetListItem extends React.Component {
     dataset: PropTypes.object,
     widget: PropTypes.object,
     layer: PropTypes.object,
-    metadata: PropTypes.array,
+    metadata: PropTypes.object,
     mode: PropTypes.string,
     user: PropTypes.object,
     tags: PropTypes.node,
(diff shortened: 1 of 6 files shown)


redactions/partnerDetail.js
```diff
@@ -125,7 +126,7 @@ export function getDatasets(ids) {
       .then((response) => {
         dispatch({
           type: GET_DATASETS_SUCCESS,
-          payload: response
+          payload: WRISerializer({ data: response })
```
mbarrenechea: return WRISerializer(response, { locale: common.locale });

components/datasets/list/list-item/component.js
```diff
@@ -27,13 +27,30 @@ class DatasetListItem extends React.Component {
     dataset: PropTypes.object,
     widget: PropTypes.object,
     layer: PropTypes.object,
-    metadata: PropTypes.object,
+    metadata: PropTypes.array,
```
mbarrenechea: Rollback

tomasmoose:
@mbarrenechea updated now !

PR #687, opened 2018-04-05
Feature/real blog posts on home

edbrett:
Updates the homepage to have REAL blog posts for the ACTUAL BLOG 😱. Following changes:
- New `<BlogLatestPosts>` component that fetches the 3 latests posts
- Reuse old styles and static card component to make the grid
- Updates styles and links for fully clickable cards
- Add loader to give user feedback

Task: https://www.pivotaltracker.com/story/show/156503368

NOTE: new `.env` variable is needed
`BLOG_API_URL=https://staging.resourcewatch.org/blog_/wp-json/wp/v2`


mbarrenechea:
Good work, let's wait for @rrequero to handle the new `/blog` url. So the `.env` variable would be ```BLOG_API_URL=https://staging.resourcewatch.org/blog_/wp-json/wp/v2```

PR #690, opened 2018-04-05
Map: thumbnail

mbarrenechea:
## Overview
This PR adds:
- Map thumbnail for dataset and widget cards
- It allows `featureService`, `mapService`, `arcgis` providers

## Still missing
- [x] support gee images @davidsingal could you work on this?


davidsingal:
I've just added:

* Support for gee images
* Fixing a bug to load basemap even when cartodb layers are failing
* Fixing styles for footer (sorry for that, I was getting an advantage while API is down)

More things still missing:

- [ ] Support for leaflet layers

davidsingal:
@mbarrenechea ready to check

PR #747, opened 2018-04-18
Log to GA the url requested when reaching a 404 page

pablopareja:


mbarrenechea:
Are you sure that we need to `logEvent`instead of `logPageView`? I think that we need to `initGA`and `logPageView`

pablopareja:
Why not `logEvent`?

PR #750, opened 2018-04-18
Map: interactivity fixed

mbarrenechea:
## Overview
This PR fixes the interactivity behavior inside the layer form. We should check that everything works

## Pivotal task
https://www.pivotaltracker.com/story/show/156876020


mbarrenechea:
It's working. We found that if we want to show the date of layers we need to cast dates to text. See this layer => https://resourcewatch.org/admin/data/layers/f3ee5110-083d-4a94-8b5d-af087d2b18aa/edit

image

mbarrenechea:
So the problem is that the layer you sent me doesn't have it. Ask data team to change it

PR #757, opened 2018-04-24
Update nighttime imagery layer id

fgassert:
## Overview
New layer ID from moving nighttime imagery layer into separate dataset.

### Test
Check that layer loads in pulse


pablopareja:
Hi Francis,
I've just checked it and it loads fine. Merging it...

PR #790, opened 2018-06-07
Feat / enable user metadata save/patch (widgets)

tomasmoose:
## Overview
https://www.pivotaltracker.com/story/show/157039831

Now user can save caption in the metadata

Moved title to the widget editor, this is so we keep it consistent with the other widget editor locations.

screen shot 2018-06-07 at 14 00 38

### DIFF
--- a/components/app/myrw/widgets/pages/edit.js
+++ b/components/app/myrw/widgets/pages/edit.js
@@ -19,16 +19,15 @@ import Input from 'components/form/Input';
 import Field from 'components/form/Field';
 
 const FORM_ELEMENTS = {
-  elements: {
-  },
+  elements: {},
   validate() {
-    const elements = this.elements;
+    const { elements } = this;
     Object.keys(elements).forEach((k) => {
       elements[k].validate();
     });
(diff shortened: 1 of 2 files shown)


components/app/myrw/widgets/pages/edit.js
```diff
@@ -95,49 +95,56 @@ class WidgetsEdit extends React.Component {
...
+
+    const { widget, name } = this.state;
     const widgetAtts = widget.attributes;
-    const dataset = widgetAtts.dataset;
+    const { dataset } = widgetAtts;
     const { user } = this.props;
     const widgetConfig = (this.onGetWidgetConfig) ? await this.getWidgetConfig() : {};
 
+    const metadata = {
+      language: this.props.locale,
```
tomasmoose: @pablopareja see there is a task here: https://www.pivotaltracker.com/story/show/156874971 

Should I always use `en`?

tomasmoose: solved ^^

PR #798, opened 2018-08-06
Add sort by relevance on keyword search

tiagojsag:
## Overview
This add the possibility to sort by "relevance" in the explore page. There's a bit of logic behind it:

1. Sort by relevance is only supported when doing a search by keyword (otherwise the API should return a 400 error)
2. Sort by relevance should be the default sort criteria when performing a keyword search
3. Sort by relevance only supports descending order

This implementation automatically switches to sorting by relevance when doing a keyword search, and falling back to the default when clearing the search keyword. It also keeps track of whether or not a user as manually set a sort criteria, in which case that takes precedence over the automatic switch to relevance sorting.

## Testing instructions
- On a fresh loaded page, doing a keyword search should automatically switch sorting to relevance
- On a session where the user previously set the sorting criteria or direction manually, automatic switching to relevance on keyword search should not occur
- If sort by relevance is set, clearing the search criteria should restore the default search criteria
- Sort by relevance should only be available if a search keyword is set, otherwise it should not show up


css/components/app/explore/explore-datasets-header.scss
```diff
@@ -33,11 +33,14 @@
         color: $color-charcoal-grey;
         fill: rgba($color-charcoal-grey, 0.2);
         display: flex;
-        cursor: pointer;
 
-        &:hover {
-          color: darken($color-charcoal-grey, 15%);
-          fill: rgba($color-charcoal-grey, 0.5);
+        &.hover {
```
davidsingal: I think we can change this class name for `isInteractive` or something like that.

PR #799, opened 2018-08-08
FEAT Subscriptions 

tomasmoose:
### Allow users to draw their own subscription area on a map

__Note__: we cant have both the country selector and Upload/Draw values active. So added a new UX element if you select a country from the list. so its clear what action is needed from the user.  Screenshot below:
screen shot 2018-08-09 at 15 56 14

__Another note__: I thought it was a good idea to show the previous selection still in **blue** on the map. This is so you can compare against your previous geo settings while drawing. Screenshot below:

screen shot 2018-08-09 at 16 30 29

Full Feature with (country geo, drawing map, upload) components:
screen shot 2018-08-09 at 15 58 05

+ Moved the upload modal out and removed the logic for toggling the modal view. as the new mockups don't require a modal window. 

+ Cleaned up styles, and renamed as the upload component is not a modal anymore. 

#### Bug fixes and improvements. 

1. Fix country selector returning null values in list

Before:
screen shot 2018-08-09 at 13 19 25

After:
screen shot 2018-08-09 at 13 18 51

2. When removing selected country area, make sure to remove geostore from state

### DIFF
--- a/components/areas/AreasForm.js
+++ b/components/areas/AreasForm.js
@@ -30,19 +30,11 @@ const MAP_CONFIG = {
     lat: 0,
     lng: 0
   },
-  zoomControl: false
-};
-
-const mapStyles = {
-  display: 'inline-block',
-  height: '410px',
-  position: 'relative',
-  width: '100%'
+  zoomControl: true
 };
 
 const FORM_ELEMENTS = {
-  elements: {
-  },
+  elements: { },
   validate() {
     const { elements } = this;
     Object.keys(elements).forEach((k) => {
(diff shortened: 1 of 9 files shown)


tomasmoose:
@mbarrenechea was a problem with the leaflet draw version we where using, pr updated

tomasmoose:
@mbarrenechea now you can draw multiple collections as well. I'm sending them in this format

```json
"geojson": {
   "type":  "FeatureCollection",
   "features": [...DrawnLayersHere]
}
```

PR #803, opened 2018-08-16
Revert "Upgrade widget editor to 1.3.6"

pablopareja:
Reverts resource-watch/resource-watch#802


pablopareja:
The upgrade was causing quite a few errors on the Explore page

PR #811, opened 2018-08-28
Explore Detail: Remove href from anchor in add to map Link

fgassert:
## Overview
A user reported that opening the Add to Map link in a new tab doesn't work (the layer doesn't show up). It looks like the Link route and anchor href differ causing inconsistent behavior.

## Test
Confirm add-to-map still works.


pablopareja:
Hi @fgassert !
I'm not able to reproduce the issue that you mention. Could you check the most updated version of the code on http://preproduction.resourcewatch.org/ and let me know if you are still experiencing the issue you mention?
Thanks in advance!

PR #816, opened 2018-08-30
use wri-api-comp map on alerts also

tomasmoose:
https://www.pivotaltracker.com/story/show/159926412

question, is zoom to fit the layer data default? seems like a thing I should not have to touch right @mbarrenechea  

atm I just default the zoom to 0.

Also, we should probably remove the map component inside RW at this point?


mbarrenechea:
You need to pass a prop called `bounds`. https://resource-watch.github.io/wri-api-components/#!/Maps

```
bounds = {{
  bbox: [],
  options: {}
}}
```

I'm going to do it

PR #845, opened 2018-10-29
Allow dot notation for raster interaction

pablopareja:
## Overview
Allow raster interactivity to be specified with dot notation. For instance:
```
"output": [
{
"column": "x.b1.mean",
"property": "Energy",
"type": "numeric",
"format": "0.00"
}
]
```

## Testing instructions
http://localhost:9000/data/explore?zoom=2&lat=-5.615985819155327&lng=0.17578125&basemap=dark&labels=light&layers=%255B%257B%2522dataset%2522%253A%2522cefd2836-3cc1-43dc-a9d1-bfecf79a6252%2522%252C%2522opacity%2522%253A1%252C%2522layer%2522%253A%25220668f151-1116-4a27-bb16-761cf259d3aa%2522%257D%255D&page=1&sort=relevance&sortDirection=-1&search=vegetation

### DIFF
--- a/components/ui/map/popup/LayerPopup.js
+++ b/components/ui/map/popup/LayerPopup.js
@@ -117,6 +117,7 @@ class LayerPopup extends React.Component {
     }
     // Get interactionConfig
     const { interactionConfig } = layer;
+    const { output } = interactionConfig;
 
     // Get data from props or state
     const interaction = layersInteraction[layer.id] || {};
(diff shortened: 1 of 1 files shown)


components/ui/map/popup/LayerPopup.js
```diff
@@ -140,17 +141,27 @@ class LayerPopup extends React.Component {
...
-                    key={outputItem.property || outputItem.column}
-                  >
-                    <td className="dt">
-                      {outputItem.property || outputItem.column}:
-                    </td>
-                    <td className="dd">{this.formatValue(outputItem, (interaction.data || interactionState.data)[outputItem.column])}</td>
-                  </tr>
-                ))}
+                {output.map((outputItem) => {
+                  const { column } = outputItem;
```
mbarrenechea: What about this. I think that using this function will prevent you to make a difference between normal ones and splitted ones
```
const columnArr = column.split('.');
const value = columnArray.reduce((acc, c) => {
  return acc[c];
}, interaction.data || interactionState.data)
```

PR #847, opened 2018-11-07
shorten long URL with bitly in share modal option

mluena:
## Overview
Shorten/hash long URLs when sharing.

## Testing instructions
Go to Data / Explore DataSets
Click on share button of map controls panel.
The URL that appears as "Public url to share" should be shorten by bitly

## Pivotal task
https://www.pivotaltracker.com/n/projects/1374154/stories/161620064.

---

[ ] Title: Don't forget to make it clear for everyone, even for people that don't know about the feature/bug.
[ ] Meaningful commits and code rebased on `develop`.


build/config.gypi
```diff
@@ -0,0 +1,72 @@
+# Do not edit. File was generated by node-gyp's "configure" step
```
mbarrenechea: Remove this file

PR #921, opened 2019-03-14
Update blog url

tiagojsag:
## Overview
resourcewatch.org/blog -> blog.resourcewatch.org

## Testing instructions
- run yarn
- notice an error about a missing protocol
- debug that
- notice that there's an undocumented env var
- document and add that env var
- run again
- yarn still complains about "TypeError: Converting circular structure to JSON"
- try to figure it out, maybe read the README
- no luck
- give up
- do the PR anyway, and tell @andresgnlez to fix it, and then test it
- @andresgnlez , while testing this, should look that all links to the blog point to blog.resourcewatch.org. Keep in mind that there's already a server level redirection in place, so check the actual html, not by clicking the links.

## Pivotal task
no

---


andresgnlez:
Seems fine to me 👌 

@tiagojsag 's issue fixed updating `.env` file.

Merging.

PR #981, opened 2019-05-03
Feature/datasets pagination

mluena:
## Overview
Add dynamic pagination in admin to fetch datasets by page.

## Testing instructions
http://localhost:9000/admin/data/datasets and change page.

## Pivotal task
https://www.pivotaltracker.com/story/show/165567043.

---

### DIFF
--- a/components/datasets/table/DatasetsTable.js
+++ b/components/datasets/table/DatasetsTable.js
@@ -82,10 +82,12 @@ class DatasetsTable extends React.Component {
   }
 
   onChangePage(page) {
+    console.log('dataser')
     this.props.getDatasets({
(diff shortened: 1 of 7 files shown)


components/datasets/table/DatasetsTable.js
```diff
@@ -58,11 +58,15 @@ class DatasetsTable extends React.Component {
     super(props);
 
     this.onSearch = this.onSearch.bind(this);
+    this.onChangePage = this.onChangePage.bind(this);
```
andresgnlez: Replace with an arrow function

andresgnlez:
@mluena 

image

wrong pagination initialization (should be 1 instead of 0)

PR #985, opened 2019-05-07
Adds pagination to layers in admin

andresgnlez:
## Overview
- Adds pagination to layers in admin zone.
- Disconnects data from the store.
- Removes several methods in the layer service replaced with standard fetches.
- Added fetch for layers and for removing layer.
- Removes `redactions/admin/layers` as we don't need it anymore.

## Testing instructions
Go to `http://localhost:9000/admin/data/layers` and play with pagination.

## Pivotal task
https://www.pivotaltracker.com/story/show/165567043

---


components/admin/data/layers/table/constants.js
```diff
@@ -0,0 +1,14 @@
...
+  limit: 20,
+  // current page
+  page: 1,
+  // total of pages
+  pages: null,
+  // total of items to display
+  size: null
+};
+
+export default { INITIAL_PAGINATION };
```
mluena: This file is the same for the three tabs (layers, widgets and datasets), maybe we should move things around and let them share it

mluena:
I get this error when I try to edit one of the layer (just clicking on edit)

Screenshot 2019-05-08 at 09 38 28

To reproduce http://localhost:9000/admin/data/layers/0699ced2-960f-4f06-8792-62f6c34b4e39/edit

maybe you don't get the error at first attempt but if you go back and press edit again you will

andresgnlez:
@mluena this is a particular issue with this layer. It's not set it up correctly.

components/admin/data/layers/table/constants.js
andresgnlez: Let's not share states. This could give us problems in the future. Keep them as independent local states.

PR #990, opened 2019-05-09
Table footer in admin restricted

mluena:
## Overview
Condition added to table footer, it doesn´t appear if there are no results.

## Testing instructions
Go to any of the options in admin (http://localhost:9000/admin/pages) and search until you get no results, table footer must disappear.

---

### DIFF
--- a/components/ui/customtable/footer/TableFooter.js
+++ b/components/ui/customtable/footer/TableFooter.js
@@ -26,16 +26,13 @@ class TableFooter extends PureComponent {
 
     return (
       <div className="table-footer">
-        {(pagination.size > 0) &&
-          <Paginator
-            options={pagination}
-            onChange={page => this.onChangePage(page)}
-          />
-        }
-        {(pagination.enabled && showTotalPages && pagination.pages && (pagination.size > 0)) &&
+        <Paginator
+          options={pagination}
+          onChange={page => this.onChangePage(page)}
+        />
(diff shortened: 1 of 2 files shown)


components/ui/customtable/footer/TableFooter.js
```diff
@@ -23,20 +23,22 @@ class TableFooter extends PureComponent {
...
-          onChange={page => this.onChangePage(page)}
-        />
-
-        {(pagination.enabled && showTotalPages && pagination.pages) &&
-          <div>Page <span>{pagination.page}</span> of <span>{pagination.pages}</span></div>
-        }
-
-      </div>
-    );
+    if (pagination.size > 0) {
```
andresgnlez: Set this condition in the table itself. It's cleaner and you don't compromise the behavior of the component.

components/ui/customtable/footer/TableFooter.js
```diff
@@ -23,20 +23,22 @@ class TableFooter extends PureComponent {
...
+          />
+
+          {(pagination.enabled && showTotalPages && pagination.pages) &&
+            <div>Page <span>{pagination.page}</span> of <span>{pagination.pages}</span></div>
+          }
+
+        </div>
+      );
+    }
+    return null;
```
andresgnlez: We haven't had good experiences with this thing. Depending on where renders it can even crash the application. Let's try to avoid it.

PR #991, opened 2019-05-09
Fix/logout error

mluena:
## Overview
Bug fixed in log out (user unable to log out in staging)

Screenshot 2019-05-09 at 15 46 15
Log out function moved to UserService,

## Testing instructions
Log in and out.

---

### DIFF
--- a/layout/header/header-user/component.js
+++ b/layout/header/header-user/component.js
@@ -5,6 +5,8 @@ import { toastr } from 'react-redux-toastr';
 import TetherComponent from 'react-tether';
 import debounce from 'lodash/debounce';
 
+import { logout } from 'services/UserService';
+
 // components
 import Icon from 'components/ui/Icon';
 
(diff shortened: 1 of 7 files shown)


layout/header/header-user/component.js
```diff
@@ -15,18 +17,13 @@ class HeaderUser extends PureComponent {
...
-    // TO-DO: move this to an action
-    fetch(`${process.env.CONTROL_TOWER_URL}/auth/logout`, { credentials: 'include' })
-      .then(() => {
-        window.location.href = `/logout?callbackUrl=${window.location.href}`;
-      })
-      .catch((err) => {
-        toastr.error('Error', err);
+  onLogout = () => {
+    logout()
+      .catch(({ response }) => {
```
andresgnlez: Not sure 100% but I don't think you receive here the response. Based on you function you will receive an error object with a `message` property. Could you double check this, please?

services/UserService.js
```diff
@@ -375,3 +378,26 @@ export default class UserService {
...
+ */
+export const logout = () => CTAPI.get('/auth/logout', {
+  headers: {
+    ...CTAPI.defaults.headers,
+    // TO-DO: forces the API to not cache, this should be removed at some point
+    'Upgrade-Insecure-Requests': 1
+  }
+}).then((response) => {
+  const { status, statusText } = response;
+  window.location.href = `/logout?callbackUrl=${window.location.href}`;
```
andresgnlez: This is part of your callback. Think about it: if I want to call this function in another side, with your code, no matter what, is going to redirect to that URL. 

I would move the redirect as part of the callback and return here just the `response` as there's no data to return.

PR #997, opened 2019-05-13
Fix/widgets list

mluena:
Overview
Bug - Re-populating datasets when you create or edit a widget.

Testing instructions
http://localhost:9000/admin/data/datasets/{ select an specific dataset }/widgets
and create a new one or edit one existing

Pivotal task
https://www.pivotaltracker.com/story/show/165864235

---

### DIFF
--- a/components/app/myrw/widgets/pages/new.js
+++ b/components/app/myrw/widgets/pages/new.js
@@ -57,24 +57,15 @@ class WidgetsNew extends React.Component {
 
     // Services
     this.widgetService = new WidgetService(null, { apiURL: process.env.WRI_API_URL });
-    this.datasetsService = new DatasetsService({
-      language: props.locale
-    });
-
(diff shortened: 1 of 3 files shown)


andresgnlez:
I'll take a look at this.

components/app/myrw/widgets/pages/new.js
```diff
@@ -162,15 +153,14 @@ class WidgetsNew extends React.Component {
     });
   }
 
-  handleChange(value) {
+  handleChange = (value) => {
     const newWidgetObj = Object.assign({}, this.state.widget, value);
     this.setState({ widget: newWidgetObj });
+    this.loadDatasets();
```
andresgnlez: This change shouldn't be in this file. Also, before anything... Could you check if with https://github.com/resource-watch/resource-watch/pull/1000 the problem is solved?

PR #1017, opened 2019-06-12
Catalog page: pagination and other improvements

pablopareja:
image

## Overview
This PR modifies the catalog page and related components so that they comply with the following: 

- server-side fetching should be removed and the front-end should be used for that instead
- datasets are fetched directly from the component and the internal component state is used to handle that data. _Redux is not used in any way for this_
- clean any code related to the Redux approach that was previously used

## Testing instructions
Go to `http://localhost:9000/catalog`

## [Pivotal task](https://www.pivotaltracker.com/story/show/164135726)


layout/app/catalog/component.js
```diff
@@ -1,42 +1,90 @@
-import React, { PureComponent } from 'react';
-import PropTypes from 'prop-types';
+import React, { Component } from 'react';
```
andresgnlez: Use always `PureComponent` unless you need to change the `shouldComponentUpdate` behaviour.

layout/app/catalog/component.js
```diff
@@ -1,42 +1,90 @@
...
-  };
+    this.state = {
+      datasets: [],
+      loading: true,
+      search: null,
+      page: 1,
+      totalItems: 0
+    };
+
+    this.handlePageChange = this.handlePageChange.bind(this);
```
andresgnlez: Use arrow function to bind the function.

layout/app/catalog/component.js
```diff
@@ -1,42 +1,90 @@
...
+    super(props);
 
-class CatalogLayout extends PureComponent {
-  static propTypes = {
-    list: PropTypes.array.isRequired,
-    loading: PropTypes.bool.isRequired,
-    getDatasets: PropTypes.func.isRequired,
-    setDatasetsSearch: PropTypes.func.isRequired
-  };
+    this.state = {
```
andresgnlez: Move the state out and get rid of the constructor.

layout/app/catalog/component.js
```diff
@@ -1,42 +1,90 @@
...
+      page: 1,
+      totalItems: 0
+    };
+
+    this.handlePageChange = this.handlePageChange.bind(this);
+  }
 
-  getSearchResults = debounce((value) => {
-    const { getDatasets } = this.props;
+  componentDidMount() {
```
andresgnlez: I think `componentWillMount` would be better, the sooner if triggers the fetch, the less the user has to wait for results.

layout/app/catalog/component.js
```diff
@@ -1,42 +1,90 @@
...
+      .then((result) => {
+        const { datasets, meta } = result;
+        this.setState({
+          datasets,
+          totalItems: meta['total-items'],
+          loading: false
+        });
+      })
+      .catch((error) => {
+        toastr.error('There was an error with the request', error);
```
andresgnlez: At this point, `error` is probably an Error object (coming from `new Error`), so if you want to render something, call its `message` property instead. Otherwise, it will crash.

layout/app/catalog/component.js
```diff
@@ -68,18 +116,22 @@ class CatalogLayout extends PureComponent {
                 </div>
               </div>
               <div className="column small-12">
-                <Spinner
-                  className="-light -relative"
-                  isLoading={loading}
-                />
+                <Spinner className="-light -relative" isLoading={loading} />
```
andresgnlez: If a component has a property, usually you keep it in the same line. More than one, you put them in different lines. Makes code easier to read.

layout/app/catalog/component.js
```diff
@@ -68,18 +116,22 @@ class CatalogLayout extends PureComponent {
...
               </div>
             </div>
 
             <div className="row">
               <div className="column small-12">
-                <DatasetList
-                  list={list}
-                  mode="list"
+                <DatasetList list={datasets} mode="list" />
```
andresgnlez: --> https://github.com/resource-watch/resource-watch/pull/1017/files#r292953618

layout/app/catalog/component.js
```diff
@@ -1,42 +1,90 @@
...
+
+    fetchDatasets(
+      {
+        'page[size]': DATASETS_PER_PAGE,
+        'page[number]': page,
+        status: 'saved',
+        published: true,
+        application: process.env.APPLICATIONS,
+        language: 'en',
+        includes: ['metadata', 'widget', 'layer', 'vocabulary'].join(','),
```
andresgnlez: Could you, please, double check if we really need all these incudes..? Checking the current site, it doesn't seem we are using most of them.

andresgnlez:
I've found a bug:
when you perform a search being in X page, the fetch params look like this: https://api.resourcewatch.org/v1/dataset?env=production,preproduction&page[size]=12&page[number]=29&status=saved&published=true&application=rw&language=en&includes=metadata,widget,layer,vocabulary&search=water&sort=name,description

Every time the user performs a search, we need to default, at least, the `page[number]` to 1, otherwise, it won't return anything.

andresgnlez:
@pablopareja the param sent for filtering by name was wrong, it's not `search` but `name`. Fixed and merging :) Good job!

PR #1018, opened 2019-06-13
Fixes fetching widgets in MyRW

andresgnlez:
## Overview
Mainly, fixes a bug in MyRW where the widgets of a user won't load.

-Removed any local filters like sorting or filtering. Now everything is done via API.
- Improved a little interface (better spinner position, heights).
- Added the possibility of retry the widget fetching if an error happens.
- Implemented React Hooks for the first time.
- Removed unnecessary selector and deprecated redaction.

## Testing instructions
Go to `http://localhost:9000/myrw/widgets/my_widgets` and play with your widgets.

## Pivotal task
https://www.pivotaltracker.com/story/show/166645058

---


components/app/myrw/widgets/tabs/list/helpers.js
```diff
@@ -0,0 +1,25 @@
...
+  } = props;
+  const { page, limit } = pagination;
+  const isCollection = !['my_widgets', 'favourites'].includes(subtab);
+
+  return ({
+    'page[size]': limit,
+    'page[number]': page,
+    sort: sort === 'asc' ? 'updatedAt' : '-updatedAt',
+    userId: id,
+    ...search && search.length && { name: search },
```
pablopareja: @andresgnlez Just a quick question, why using the spread operator here instead of the variable `search` directly ?

andresgnlez: It's a way to conditionally add properties to the object. I check previously if the search value is right, otherwise, we would always be sending search empty to the API (which is fine, it doesn't affect the results, but I think this is more cleaner).

pablopareja: Oh I see now. Thanks for the clarification 👍

PR #1020, opened 2019-06-14
MyRW fix: redundant "Explore datasets" button removed from Datasets section

pablopareja:
image

## Overview
This PR removes a redundant "Explore datasets" button appearing in the Myrw --> Datasets section

## Testing instructions
Go to `http://localhost:9000/myrw/datasets/my_datasets`

## [Pivotal task](https://www.pivotaltracker.com/story/show/166694334)

---

### DIFF
--- a/components/app/myrw/datasets/pages/my-rw-datasets/dataset-list/dataset-list-component.js
+++ b/components/app/myrw/datasets/pages/my-rw-datasets/dataset-list/dataset-list-component.js
@@ -18,7 +18,10 @@ class DatasetsList extends PureComponent {
     },
     datasets: [],
     filters: [],
-    loading: true
+    loading: true,
+    user: {},
+    currentTab: '',
+    getDatasetsByTab: () => null
   };
 
   static propTypes = {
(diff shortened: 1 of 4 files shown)


components/app/myrw/datasets/pages/my-rw-datasets/dataset-list/dataset-list-component.js
```diff
@@ -18,7 +18,10 @@ class DatasetsList extends PureComponent {
     },
     datasets: [],
     filters: [],
-    loading: true
+    loading: true,
+    user: {},
+    currentTab: '',
+    getDatasetsByTab: () => null
```
andresgnlez: Let's try to avoid this kind of things. Usually, this kind of functions are mandatory so better adding an `isRequired` attribute and avoids confusion.

In case of a function is really not required, better asking in the code if it exists instead of executing it.

components/app/myrw/datasets/pages/my-rw-datasets/dataset-list/dataset-list-component.js
```diff
@@ -47,7 +45,7 @@ class DatasetsList extends PureComponent {
 
     toastr.confirm(
       `Are you sure you want to delete the dataset: ${
-      metadata && metadata.attributes.info ? metadata.attributes.info.name : dataset.name
+        metadata && metadata.attributes.info ? metadata.attributes.info.name : dataset.name
```
andresgnlez: We would need to update the fetching of this component... It seems is still using the deprecated service as we have `attributes` here 🙄.  I'll add a task on PT: https://www.pivotaltracker.com/story/show/166727155

PR #1024, opened 2019-06-19
Improvements for MyRWDatasets and DatasetList components

pablopareja:
## Overview
This PR includes the following fixes and improvements:

- The component uses now the right service `services/dataset` instead of the old approach to handle datasets fetching and deletion as part of the MyRW --> datasets section
- WRISerializer is used now to convert the results from the requests to the appropriate JSON format
- Pagination toolbar is only shown when there are two or more pages of data
- Fix for a bug that prevented successfully loading the user datasets after one dataset had been removed 

## Testing instructions
Go to `http://localhost:9000/myrw/datasets/my_datasets`


andresgnlez:
@pablopareja feel free to merge.

PR #1028, opened 2019-06-24
Fix back office navigation on save

pablopareja:
## Overview
This PR fixes the navigation behavior when clicking “save” button after creating/editing both layers and widgets. Now the app stays on same page rather than navigating back to the list of layers or widgets.

## Testing instructions
Go to `http://localhost:9000/admin/data/layers/` and `http://localhost:9000/admin/data/widgets/` and create + edit layers and widgets.

## [Pivotal task](https://www.pivotaltracker.com/story/show/166845219)

### DIFF
--- a/components/admin/data/widgets/pages/show/component.js
+++ b/components/admin/data/widgets/pages/show/component.js
@@ -18,9 +18,9 @@ class WidgetsShow extends PureComponent {
   handleSubmit = (widget) => {
     if (widget) {
       Router.pushRoute('admin_data_detail', {
-        tab: 'datasets',
-        subtab: 'widgets',
-        id: widget.dataset
+        tab: 'widgets',
+        id: widget.id,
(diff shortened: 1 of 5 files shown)


components/admin/data/widgets/pages/show/component.js
```diff
@@ -18,9 +18,9 @@ class WidgetsShow extends PureComponent {
   handleSubmit = (widget) => {
     if (widget) {
       Router.pushRoute('admin_data_detail', {
```
andresgnlez: Aren't we just trying to go to the same page after updating a widget? I think we could do nothing for that case.

pablopareja: @andresgnlez I've just updated the code so that the page scrolls to the top instead of reloading it.
Should we merge it now?

PR #1032, opened 2019-06-28
Clean up of obsolete/unused images plus Splash page related code

pablopareja:
## Overview
This PR removes all images found in static that are not currently used. 
All the code related to the Splash + Splash detail pages has also been removed.


constants/app.js
```diff
@@ -11,13 +11,11 @@ export const TRANSIFEX_BLACKLIST = [
 ];
 
 export const USERREPORT_BLACKLIST = [
```
andresgnlez: Let's remove this constant if we are not using it anymore.

andresgnlez: @pablopareja I did it myself (https://github.com/resource-watch/resource-watch/pull/1032/commits/f68a8fc12b7b83ead081513a26647e9c98bd8ab1), please check and merge if everything is fine 👍

PR #1039, opened 2019-07-09
Handle icons that are RW specific using Icomoon directly from the RW app

pablopareja:
## Overview
This PR creates the infrastructure to handle RW specific icons with Icomoon directly from the RW app.

## [Pivotal task](https://www.pivotaltracker.com/story/show/167088413)


layout/layout/layout-app/component.js
```diff
@@ -9,6 +9,8 @@ import { browserSupported } from 'utils/browser';
 
 // vizzuality-components
 import { Icons } from 'vizzuality-components';
+// RW icons
```
andresgnlez: They're not unique or special, just components :P

andresgnlez:
@pablopareja are we missing importing the RW icons in the `admin` layout?

PR #1047, opened 2019-07-11
Layer info pop-up "more info" button leads to `resourcewatch.org/data/explore/[slug]`

pablopareja:
## Overview
This PR updates the "more info" button from the layer info pop-up to direct to `resourcewatch.org/data/explore/[slug]` instead of `resourcewatch.org/data/explore/[id]`

## Testing instructions
Add any dataset to the Explore page and open the layer info pop up. The "More info" button should link to the dataset detail page using the slug instead of the dataset id.

## [Pivotal task](https://www.pivotaltracker.com/story/show/167211371)

### DIFF
--- a/components/modal/layer-info-modal/layer-info-modal-component.js
+++ b/components/modal/layer-info-modal/layer-info-modal-component.js
@@ -7,9 +7,6 @@ import { connect } from 'react-redux';
 
 import { fetchDataset } from 'services/dataset';
 
-// Styles
-import './styles.scss';
-
 function LayerInfoModal(props) {
   const { embed, layer } = props;
   const [slug, setSlug] = useState(' ');
@@ -26,7 +23,7 @@ function LayerInfoModal(props) {
       <div className="layer-info-content">
         <h2>{layer.name}</h2>
         <p>{layer.description}</p>
-        <div className="buttons">
+        <div className="c-button-container -j-end">
           {embed && (
             <a
               className="c-btn -primary"
--- a/components/modal/layer-info-modal/layer-info-modal-component.js
+++ b/components/modal/layer-info-modal/layer-info-modal-component.js
@@ -2,19 +2,16 @@ import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'routes';
 
-// Redux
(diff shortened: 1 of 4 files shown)


components/modal/layer-info-modal/layer-info-modal-component.js
```diff
@@ -1,55 +1,57 @@
...
-          </div>
+import { fetchDataset } from 'services/dataset';
+
+function LayerInfoModal(props) {
+  const { embed, layer } = props;
+  const [slug, setSlug] = useState(' ');
+
+  useEffect(() => {
+    fetchDataset(layer.dataset)
+      .then((response) => {
```
andresgnlez: Let's make thing easier: rename to `dataset`

components/modal/layer-info-modal/layer-info-modal-component.js
```diff
@@ -1,55 +1,57 @@
...
-                  More info
-                </a>
-              </Link>
-            }
-          </div>
+import { fetchDataset } from 'services/dataset';
+
+function LayerInfoModal(props) {
+  const { embed, layer } = props;
+  const [slug, setSlug] = useState(' ');
```
andresgnlez: `(' ')` what's this?

andresgnlez:
@pablopareja could you move the styles of the layer info modal to component-level? Also, taking I look at the styles... I think we can get rid of them using the styles for `c-button-container`

components/modal/layer-info-modal/layer-info-modal-component.js
```diff
@@ -1,55 +1,57 @@
...
+  }, []);
+
+  return (
+    <div className="layer-info-modal">
+      <div className="layer-info-content">
+        <h2>{layer.name}</h2>
+        <p>{layer.description}</p>
+        <div className="buttons">
+          {embed && (
+            <a
```
andresgnlez: Let's use `Link`.

components/modal/layer-info-modal/layer-info-modal-component.js
pablopareja: If we use the empty string instead, there's the following error thrown when the `<Link>` is built as part of the render:
image

PR #1075, opened 2019-07-26
Fix for: "Array error when saving widget"

pablopareja:
## [Pivotal task](https://www.pivotaltracker.com/story/show/167518460)

### DIFF
--- a/services/WidgetsService.js
+++ b/services/WidgetsService.js
@@ -47,6 +47,7 @@ export default class WidgetsService {
         type,
         body: {
           ...body,
+          ...type === 'PATCH' && { application: [body.application] },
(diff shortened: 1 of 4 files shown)


services/WidgetsService.js
```diff
@@ -76,7 +77,8 @@ export default class WidgetsService {
         type,
         body: {
           ...body,
-          application: process.env.APPLICATIONS
+          ...type === 'PATCH' && { application: [body.application] },
```
pablopareja: @andresgnlez This line is redundant

services/WidgetsService.js
```diff
@@ -47,7 +47,8 @@ export default class WidgetsService {
...
```
pablopareja: @andresgnlez This line is redundant

PR #1089, opened 2019-08-06
Alert preview improvements

pablopareja:
image

## Overview
This PR adds a disclaimer text to the preview subscription table.

## Testing instructions
Go to MyRW --> Areas of interest. Then, in the process of creating a new subscription, click on the button 'Preview'

## [Pivotal task](https://www.pivotaltracker.com/story/show/167510628)

### DIFF
--- a/components/modal/subscriptions-modal/subscriptions-preview/table/component.js
+++ b/components/modal/subscriptions-modal/subscriptions-preview/table/component.js
@@ -1,6 +1,8 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 
+import './styles.scss';
+
 class AlertsTable extends PureComponent {
   static propTypes = {
     alerts: PropTypes.shape({
@@ -17,6 +19,9 @@ class AlertsTable extends PureComponent {
     return (
       <div className="c-subscription-preview-table">
         <h3>{title}</h3>
+        <div className="disclaimer-container">
+          This is an example of what an alert subscription might look like. The information here is intended as an example only and does not represent the current status of this alert in your area of interest.
+        </div>
         {!data.length && (
           <span>Data no available</span>
         )}
--- a/components/modal/subscriptions-modal/subscriptions-preview/table/styles.scss
+++ b/components/modal/subscriptions-modal/subscriptions-preview/table/styles.scss
@@ -3,7 +3,7 @@
     width: 100%;
     background-color: rgba($yellow, 0.1);
     border-radius: 4px;
-    padding: 10px;
-    margin-bottom: 15px;
+    padding: $space * 2;
+    margin-bottom: $space * 3;
   }
 }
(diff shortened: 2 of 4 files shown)


components/modal/subscriptions-modal/area/component.js
```diff
@@ -82,6 +82,9 @@ class AreaSubscriptionsModal extends PureComponent {
     } = this.props;
     const { showSubscribePreview } = this.state;
 
+    console.log('hey!', this.props);
```
andresgnlez: Hey! Remove me ;)

components/modal/subscriptions-modal/subscriptions-preview/table/component.js
```diff
@@ -17,6 +19,9 @@ class AlertsTable extends PureComponent {
     return (
       <div className="c-subscription-preview-table">
         <h3>{title}</h3>
+        <div className="disclaimer-container">
```
andresgnlez: Do you know if this kind of disclaimer is something we're going to spread in the app? Otherwise, we could isolate it in a component. If not, leave it as it is.

components/modal/subscriptions-modal/subscriptions-preview/table/styles.scss
```diff
@@ -0,0 +1,9 @@
+.c-subscription-preview-table {
+  .disclaimer-container {
+    width: 100%;
+    background-color: rgba(250,183,46, 0.1);
```
andresgnlez: Is this a new color? Don't we have a similar? Sorry for being paranoid with the colors but you know how it can scale if we have no control over this...

components/modal/subscriptions-modal/subscriptions-preview/table/styles.scss
```diff
@@ -0,0 +1,9 @@
...
+    border-radius: 4px;
+    padding: 10px;
```
andresgnlez: Let's try to use the `$space` variable instead and keep consistency with spacing.

components/modal/subscriptions-modal/area/component.js
pablopareja: hahaha, done! xD

components/modal/subscriptions-modal/subscriptions-preview/table/component.js
pablopareja: Hmm That's a good question, I don't think we are right now, but let's take it into account and, in the case where we need to put a similar one again, we can create a component for this

components/modal/subscriptions-modal/subscriptions-preview/table/styles.scss
pablopareja: I actually searched for it before adding it explicitly but couldn't find any occurrences in the code. We could anyway create a CSS variable just in case it's needed in the future. _I'll do that in the next commit._

PR #1106, opened 2019-08-23
datasets from collections should link to different pages depending on user and ownership

pablopareja:
## Overview
This PR improves the links dataset cards are pointing to when visualizing the datasets included in a user collection in MyRW.

## Testing instructions
Go to `http://localhost:9000/myrw/collections`, create a collection that includes both datasets that you own and public ones. Check that the cards are pointing to the right pages according to the specification included in the PT tasks linked below.
_This feature should ideally be tested both by a `"normal"` user account and an `admin` account._

## [Pivotal task](https://www.pivotaltracker.com/story/show/167509722)

_PS: The `delete` button is only shown for normal users on purpose since it makes more sense for admins to manage contents directly from the back office_


andresgnlez:
@pablopareja I think same way should be applied to `WidgetCard` component as the user is also able to save a widget in a collection.

PR #1107, opened 2019-08-23
Expandable table of contents in FAQs

pablopareja:
image

## Overview
This PR adds the possibility to have expandable tables of contents in the FAQs page.

## Testing instructions
Go to `http://localhost:9000/about/faqs`


andresgnlez:
@pablopareja let's better create an `Accordion` component that allows any props and children. This way we scoped everything related to the accordion and we can use it in other places in the future if needed.

PR #1145, opened 2019-09-26
Implement update to effectively separate admin from user content in the back office

andresgnlez:
image

## Overview
This PR adds a radio button to the back office that allows to load separately content from users with roles `ADMIN` and `USER`.

## Testing instructions
Go to the back office and check the tables for:
- dataset
- widget
- layer
- dashboards

## [Pivotal task](https://www.pivotaltracker.com/story/show/167344998)


andresgnlez:
image

Let's trigger a `toastr` instead. It's more consistent.

components/admin/data/widgets/table/component.js
```diff
@@ -35,20 +36,30 @@ class WidgetsTable extends PureComponent {
     pagination: INITIAL_PAGINATION,
     loading: true,
     widgets: [],
-    filters: { name: null }
+    filters: { name: null, 'úser.role': 'ADMIN' }
```
andresgnlez: `user.role` not `úser.role` 😛 Are you testing me? 😂

andresgnlez:
In Datasets, after clicking on "user content" and trigger the error mentioned above, I click on "admin content" option again and it performs th request but it doesn't load any data, could you, please, check this behavior?

andresgnlez:
Please, add spinners every time we request new data. Right now the user doesn't know what's going on.

andresgnlez:
`https://api.resourcewatch.org/v1/layer?env=production&application=rw&includes=user&page[number]=1&page[size]=20&user.role=ADMIN` returns some layers without a user linked, how do we know they are really created by an admin user? 🤔

components/admin/data/layers/table/component.js
```diff
@@ -36,51 +37,29 @@ class LayersTable extends PureComponent {
     pagination: INITIAL_PAGINATION,
     loading: true,
     layers: [],
-    filters: { name: null }
+    filters: { name: null, 'user.role': 'ADMIN' }
   }
 
   componentDidMount() {
```
andresgnlez: Can we change it to `componentWillMount`? This way the request will fetch sooner.

components/admin/data/layers/table/component.js
```diff
@@ -89,101 +68,39 @@ class LayersTable extends PureComponent {
       filters: {
         ...filters,
         name: value
```
andresgnlez: I know you didn't touch this, but let's not send the `name` field if the value is empty. This way we avoid unexpected behaviors from the API.

components/admin/table-filters/styles.scss
```diff
@@ -0,0 +1,17 @@
+@import 'css/settings';
+
+.c-filters {
+  display: flex;
+
+  .c-radio-box {
+    display: flex;
+    .c-radio:not(:last-child) {
+      margin-right: $margin-size-extra-small;
```
andresgnlez: `$space * 3` better please, we want to get rid of this "atomic" variables.

components/admin/data/layers/table/component.js
pablopareja: @andresgnlez I tried to follow that approach but it's not possible. Consider for instance the case where the user performs a search such as "food" and then deletes the text in order to see all results. The state would keep having `name: food` if we only add the name property with a condition such as `...(value !== '' && { name: value })`

components/admin/data/widgets/table/component.js
pablopareja: you spotted it! 🔍 😅

PR #1156, opened 2019-10-03
Subscriptions review and refactor to comply with agreed approach

pablopareja:
## Overview
This PR makes the modifications needed so that subscriptions are handled with the agreed approach. That's to say, one subscription entity should be created in the API for each combination of:
- `area` 
- `dataset` 
- `subscription type`

## Testing instructions
Create/update/delete subscriptions and areas from MyRW --> AOI, Explore detail, and Planet Pulse

## Pivotal tasks
- https://www.pivotaltracker.com/story/show/168838896 (related to this)


andresgnlez:
I've found a buggy case:
 
- Create a new area.
- Add a new subscriptions. Click on "view my subscriptions" button.

The modal will close, showing the areas, but the modified area doesn't display the new subscriptions. (it fixes if I reload the page)

andresgnlez:
Another case: 

After updating any subscription in the modal, it doesn't display the new state of the subscription in the card.

Steps to follow:
- Open subscriptions modal of any area and update a subscription.
- Close the modal. Check subscription status in the card.

This bug could be related to the above 👆 one.

andresgnlez:
image

I should be able to remove any subscription.

andresgnlez:
- Open subscriptions modal with many subscriptions.
- Delete some of them. Click on "update".
- Open again the same modal. Subscriptions will be still there.

If you refresh, it's fixed.

I think there's an issue when updating the user selection in relation to the info we have of the suer.

andresgnlez:
image

The area's selector doesn't display the name of the area/country.

andresgnlez:
Clicking on the "update" button of the modal doesn't close the modal after updating the subscriptions.

andresgnlez:
@pablopareja good job 👌

PR #1162, opened 2019-10-14
Feature/refactor dataset service

pablopareja:
## Overview
This PR includes the following updates:
1. refactor to merge all the dataset services into the one located in `services/dataset`, _(we used to have three: `DatasetService`, `DatasetsService`, and `dataset`)_
2. all endpoints included in the `dataset` service use axios now and include both logging and basic error handling
3. The `graph` service has been refactored so that it's handled with axios and also includes logging and basic error handling now.
4. fix for the `TagsForm` component
5. Eslint improvement in the various files that have been updated as part of the PR

## Testing instructions
Check all features involving requests to the graph and dataset endpoints, including both the app and the back office.

## [Pivotal task](https://www.pivotaltracker.com/story/show/169012051)


components/admin/data/layers/form/LayersForm.js
```diff
@@ -146,7 +147,7 @@ class LayersForm extends PureComponent {
       if (valid) {
         // Start the submitting
         this.setState({ submitting: true });
-        this.props.dispatch(setLayerInteractionError(false));
+        this.propssetLayerInteractionError(false);
```
andresgnlez: `this.props.setLayerInteractionError`

components/subscriptions/SubscriptionCard.js
```diff
@@ -28,34 +25,26 @@ const MAP_CONFIG = {
...
-    this.datasetService = new DatasetService(props.subscription.attributes.datasetsQuery[0].id, {
-      apiURL: process.env.WRI_API_URL,
-      language: props.locale
-    });
-  }
+  state = {
+    loading: false,
+    dataset: null,
+    country: null,
+    type: this.props.subscription.attributes.datasetsQuery[0].type
```
andresgnlez: `this.props.subscription.attributes` should be `this.props.subscription`

components/subscriptions/SubscriptionCard.js
```diff
@@ -28,34 +25,26 @@ const MAP_CONFIG = {
...
+    const { subscription } = this.props;
+    const datasetId = subscription.attributes.datasetsQuery[0].id;
     this.setState({ loading: true });
-    this.datasetService.fetchData()
+    fetchDataset(datasetId)
       .then((response) => {
         const dataset = response;
         this.setState({ dataset });
-        const paramsObj = this.props.subscription.attributes.params;
+        const paramsObj = subscription.attributes.params;
```
andresgnlez: `subscription.params`

components/subscriptions/SubscriptionCard.js
```diff
@@ -28,34 +25,26 @@ const MAP_CONFIG = {
...
+    type: this.props.subscription.attributes.datasetsQuery[0].type
+  };
 
   componentDidMount() {
     this.loadData();
   }
 
   loadData() {
+    const { subscription } = this.props;
+    const datasetId = subscription.attributes.datasetsQuery[0].id;
```
andresgnlez: subscription.datasetsQuery

services/graph.js
```diff
@@ -1,136 +1,227 @@
...
+import WRISerializer from 'wri-json-api-serializer';
+
+// utils
+import { WRIAPI } from 'utils/axios';
+import { logger } from 'utils/logs';
+
+/**
+ * Get all tags
+ */
+export const fetchAllTags = (params = {
```
andresgnlez: I don't think it's a good idea having default params, you can override them accidentally at any moment and you still need them.

services/graph.js
```diff
@@ -1,136 +1,227 @@
...
+  limit: 6
+}) => {
+  logger.info('Fetch similar datasets');
+  const endpoint = withAncestors ? 'similar-dataset-including-descendent' : 'similar-dataset';
+  return WRIAPI.get(
+    `graph/query/${endpoint}`,
+    {
+      params: {
+        ...params,
+        dataset: datasetIds.join(',')
```
andresgnlez: Should be part of the `params` object. The fewer operation the service has to perform, the better.

services/graph.js
```diff
@@ -1,136 +1,227 @@
...
+      },
+      headers: { 'Upgrade-Insecure-Requests': 1 }
+    }
+  )
+    .then(response => response.data.data)
```
andresgnlez: Use Axios' `trasnsformResponse` before.

services/graph.js
```diff
@@ -1,136 +1,227 @@
...
+ * @param {string} [token] User token
+ * @returns {Promise<void>}
+ */
+export const countDatasetView = (datasetId, token, params = {
+  env: process.env.API_ENV,
+  application: process.env.APPLICATIONS
+}) => {
+  logger.info('Count dataset view');
+  return WRIAPI.post(`graph/dataset/${datasetId}/visited`,
+    {},
```
andresgnlez: Shouldn't  `params` object go here?

pablopareja: Not really, it also caught my attention when I was refactoring it. However it seems that it's a `POST` endpoint that doesn't expect a body _(I think that the only information that it needs is gathered from the url itself and the user that is performing it)_

components/admin/data/layers/form/LayersForm.js
```diff
@@ -63,22 +64,18 @@ class LayersForm extends PureComponent {
...
+        this.props.setLayerInteractionError(valid ? false : err);
       }
     });
   }
 
   componentDidMount() {
     const { id } = this.state;
 
-    const promises = [this.datasetsService.fetchAllData({})];
+    const promises = [fetchDatasets({ 'page[size]': 999999 })];
```
andresgnlez: One day will need to change this kind of behavior for fetching by demand or similar... It's kinda crazy to bring everything at once... Just pointing this 👍

services/graph.js
```diff
@@ -189,20 +195,19 @@ export const fetchMostFavoritedDatasets = (params = {
...
   logger.info('Fetch similar datasets');
   const endpoint = withAncestors ? 'similar-dataset-including-descendent' : 'similar-dataset';
   return WRIAPI.get(
     `graph/query/${endpoint}`,
     {
       params: {
-        ...params,
-        dataset: datasetIds.join(',')
+        dataset: datasetIds.join(','),
+        published: true,
```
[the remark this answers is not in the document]
pablopareja: Fixed in this commit 8767b38ac78b950be607f180fcca5d1c06177e9b

components/app/myrw/widgets/tabs/edit.js
```diff
@@ -251,8 +239,6 @@ const mapStateToProps = state => ({
   locale: state.common.locale
 });
 
-const mapDispatchToProps = dispatch => ({
-  setDataset: dataset => dispatch(setDataset(dataset))
-});
+const mapDispatchToProps = dispatch => ({ setDataset: dataset => dispatch(setDataset(dataset)) });
```
andresgnlez: Please, change to `{ setDataset }`.

components/app/myrw/widgets/tabs/new.js
```diff
@@ -134,26 +132,25 @@ class WidgetsNew extends React.Component {
...
+      .then((response) => {
+        this.setState({
+          datasets: [...this.state.datasets, ...response.map((dataset) => {
+            const metadata = dataset.metadata[0];
+            return ({
+              id: dataset.id,
+              type: dataset.type,
+              provider: dataset.provider,
+              tableName: dataset.tableName,
+              label: metadata && metadata.attributes.info
```
andresgnlez: ` metadata.info`

redactions/widget.js
```diff
@@ -124,18 +125,9 @@ export default function (state = initialState, action) {
...
-    const datasetService = new DatasetService(datasetId, {
-      apiURL: process.env.WRI_API_URL,
-      language: state.common.locale
-    });
-
-    return datasetService.fetchData('metadata')
-      .then(dataset => dispatch({ type: SET_WIDGET_DATASET, payload: dataset }));
-  };
-}
+const fetchDataset = datasetId =>
```
andresgnlez: For consistency, this should be `getDataset`. We reserve the `fetch-` for functions that perform requests to services. And you solve the naming issue with `fetchDatasetService ` ;)

layout/explore-detail/explore-detail-actions.js
```diff
@@ -93,12 +95,8 @@ export const fetchTags = createThunkAction('WIDGET-DETAIL/fetchTags', () => (dis
...
   const { exploreDetail, user } = getState();
-
-  const service = new GraphService();
-
   if (!user.token) {
     return;
   }
-
-  service.countDatasetView(exploreDetail.data.id, user.token);
+  countDatasetView(exploreDetail.data.id, user.token);
```
andresgnlez: For consistency, I would name it `fetchDatasetCount` or similar... The current name doesn't give you any hint on what it does and it looks like more a component name to me. Thoughts?

layout/explore-detail/explore-detail-actions.js
```diff
@@ -3,7 +3,10 @@ import { createAction, createThunkAction } from 'redux-tools';
 import WRISerializer from 'wri-json-api-serializer';
```
andresgnlez: In this same file, you haven't changed this https://github.com/resource-watch/resource-watch/blob/b7b8c45613e8215ca2907cb99d663a9fc88bb6bd/layout/explore-detail/explore-detail-actions.js#L24, any reason to not do it? Or just forgot about it?

services/dataset.js
```diff
@@ -13,8 +13,7 @@ import { logger } from 'utils/logs';
  * @returns {Object[]} array of serialized datasets.
  */
 export const fetchDatasets = (params = {}, headers = {}, _meta = false) => {
-  logger.info('fetches datasets');
-
+  logger.info('Fetch datasets');
```
andresgnlez: _hates you_

layout/explore-detail/explore-detail-actions.js
pablopareja: fixed here: 
17c2178

services/graph.js
```diff
@@ -1,136 +1,230 @@
...
+      throw new Error(`Error fetching all tags: ${status}: ${statusText}`);
+    });
+};
+
+/**
+ * Get inferred tags
+ */
+export const fetchInferredTags = (tags, params = {}) => {
+  logger.info('Fetch inferred tags');
+  return WRIAPI.get(`graph/query/concepts-inferred?concepts=${tags}`,
```
andresgnlez: Why `concepts ` is not part of the `params` object?

services/graph.js
```diff
@@ -1,136 +1,230 @@
...
+      const { status, statusText } = response;
+      logger.error(`Error fetching inferred tags: ${tags} ${status}: ${statusText}`);
+      throw new Error(`Error inferred tags: ${tags} ${status}: ${statusText}`);
+    });
+};
+
+/**
+* Get dataset tags
+*/
+export const fetchDatasetTags = (datasetId, params = {}) => {
```
andresgnlez: I think all tag-related endpoints should go to the `dataset` service. All endpoints start with `/dataset` which is basically the way we organize the services: under the same root endpoint. And this case is not different, we are updating the tags of a dataset.

services/graph.js
```diff
@@ -1,136 +1,230 @@
...
+      const { status, statusText } = response;
+      logger.error(`Error fetching dataset tags ${datasetId}: ${status}: ${statusText}`);
+      throw new Error(`Error fetching dataset tags ${datasetId}: ${status}: ${statusText}`);
+    });
+};
+
+/**
+* Update dataset tags
+*/
+export const updateDatasetTags = (datasetId, tags, token, usePatch = false) => {
```
andresgnlez: I think this kind of function is pure evil. This one is able to trigger 3 different fetches based on some params. I won't enter too much in detail if we are going to rebuild the graph service, but this is the kind of flow we're running away.

simaob:
Hey @andresgnlez have you had the chance to look at this again? Tks!

andresgnlez:
Looks good to me 👍

sonufrienko:
Do we really need to use `componentWillMount` instead of `componentDidMount`?
`componentWillMount` is deprecated and not recommended as an event to fetch data.

- [componentWillMount() vs componentDidMount()](https://dev.to/torianne02/componentwillmount-vs-componentdidmount-5f0n)
- [Legacy Lifecycle Methods](https://reactjs.org/docs/react-component.html#legacy-lifecycle-methods)

sonufrienko:
Because of `componentWillMount` we have a lot of warning like this

> Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.
> 
> * Move code with side effects to componentDidMount, and set initial state in the constructor.
> * Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

pablopareja:
Hi @sonufrienko, thanks for your messages 👍 
I'd say that the rationale behind our approach was to perform the API request as soon as possible so that we could minimize the data loading time.
Perhaps we should reconsider it and do it in the way you're proposing it.
Thanks for the suggestion 👌

PR #1166, opened 2019-10-31
Set origin parameter when signing up user on the API

tiagojsag:
https://www.pivotaltracker.com/story/show/168861757

I didn't test this, so you may want to.


andresgnlez:
@tiagojsag Isn't the `apps: [process.env.APPLICATIONS]` in the POST params enough to achieve this?

tiagojsag:
no

PR #1196, opened 2020-01-03
replace dataset ids of unpublised datasets

amsnyder:
## Overview
Replace dataset ids of datasets that have been unpublished and replaced with new versions of the dataset.

Removed asylum requests dataset because we have archived that dataset. It is unpublished and will no longer be maintained.

---


pablopareja:
Closing this PR since the changes will be applied here: https://github.com/resource-watch/resource-watch/pull/1198

PR #1213, opened 2020-01-28
Fix: set the default layer on open in map click

dancer2090:
## Overview
This PR fixes the selection of default layer when user click Open in Map button.
## Testing instructions
Go to http://localhost:9000/data/explore
Select a dataset with multiple layers. Go to details page. Click open in map.You will see the default layer.
## [Pivotal task](https://www.pivotaltracker.com/n/projects/1374154/stories/169711432)

---


layout/explore-detail/explore-detail-buttons/explore-detail-buttons-component.js
```diff
@@ -102,12 +102,12 @@ class ExploreDetailButtons extends PureComponent {
...
                     opacity: 1,
                     visible: true,
-                    layer: dataset.layer[0].id
+                    layer: dataset.layer.find(l => l.default).id
                   }]))
                 }}
               >
                 <a className="c-button -primary">
-                  Open in map
+                  Open in map 1
```
pablopareja: The button text should remain the same

PR #1253, opened 2020-02-18
Explore detail component plus header buttons

pablopareja:
image

## Overview
This PR includes a first draft of the Explore detail component integrated into the sidebar. The header including the buttons has been included as well as the dataset title. The dataset info is loaded by means of a hook using `userReducer` and `useEffect`.

## Testing instructions
Go to `localhost:9000/data/explore` and verify that the sidebar shows the Explore detail component when clicking on one of the dataset cards `Details` button.

### DIFF
--- a/layout/explore/explore-detail/component.js
+++ b/layout/explore/explore-detail/component.js
@@ -10,9 +10,11 @@ import './styles.scss';
 
 class ExploreDetailComponent extends React.Component {
   static propTypes = {
-    dataset: PropTypes.object.isRequired,
+    dataset: PropTypes.object,
     loading: PropTypes.bool.isRequired
   };
+
+  static defaultProps = { dataset: null };
 
 
   render() {
--- a/layout/explore/explore-detail/component.js
+++ b/layout/explore/explore-detail/component.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// Components
+import ExploreDetailHeader from './explore-detail-header';
+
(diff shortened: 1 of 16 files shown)


layout/explore/explore-detail/component.js
```diff
@@ -0,0 +1,38 @@
...
+// Components
+import Spinner from 'components/ui/Spinner';
+import ExploreDetailHeader from './explore-detail-header';
+
+// Styles
+import './styles.scss';
+
+class ExploreDetailComponent extends React.Component {
+  static propTypes = {
+    dataset: PropTypes.object.isRequired,
```
tomasmoose: Getting this error when entering the details component 

`Warning: Failed prop type: The prop dataset is marked as required in ExploreDetailComponent, but its value is null`

if it can be null it should probably not have the `isRequired` defined or make sure it's not sent as `null`

tomasmoose:
When refreshing the browser window the explore details is not remembered, I get sent back to `all datasets` is this expected @pablopareja ?

layout/explore/explore-detail/component.js
pablopareja: Fixed here: https://github.com/resource-watch/resource-watch/pull/1253/commits/140357cf63e93432504a902b5ef7eea870a4a441

pablopareja:
@tomasmoose Regarding your last comment, that's something that will be tackled in another PR once the new URL specification is 100% clear. 
Thanks for bringing it up anyway! 👍

PR #1256, opened 2020-02-21
Progress on Explore detail metadata

pablopareja:
image

## Overview
This PR includes the following changes for Explore detail:
- Main sections added
- Structure, main styles, and new components including `Further information` and `Footer`
- `Read more` functionality added as well as Markdown support for metadata fields

## Testing instructions
Go to Explore and select one dataset to open the Explore detail view on the sidebar.


tomasmoose:
Screenshot 2020-02-24 at 09 24 33

Maybe it's my OCD, but this "read more", "read less" button has weird padding to the left, either we can use a secondary button with a border or remove the left padding here?

tomasmoose:
Otherwise tested in FF, Chrome, brave, edge works great

pablopareja:
@tomasmoose Thanks for your review! 👏 
I've fixed the unnecessary margin for the `Read more` button here: https://github.com/resource-watch/resource-watch/pull/1256/commits/bcdd152cbf308798b907a54fecf688371e30fe8a
Regarding your other comment, those two sections should always have data, or at least in theory. Let's leave it like that for now so that people can see how they would look like and we can tackle what you suggested later if necessary 😉

PR #1257, opened 2020-02-24
Explore detail: dataset layers section

pablopareja:
image

## Overview
This PR adds the section `Dataset layers` to Explore detail.

## Testing instructions
Try toggling different layers on and off in different scenarios such as:
- No datasets added to the map
- Other datasets have been added to the map
- Toggle on and off layers from the same dataset
- ....

### DIFF
--- a/layout/explore/explore-detail/dataset-layers/dataset-layer-card/component.js
+++ b/layout/explore/explore-detail/dataset-layers/dataset-layer-card/component.js
@@ -1,16 +1,61 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
 
 // Styles
 import './styles.scss';
 
 function DatasetLayerCard(props) {
-  const { layer } = props;
+  const {
+    dataset,
+    layer: { id, name, description },
+    layerGroups,
+    setMapLayerGroupActive,
+    toggleMapLayerGroup
+  } = props;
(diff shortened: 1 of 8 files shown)


layout/explore/explore-detail/dataset-layers/dataset-layer-card/component.js
```diff
@@ -0,0 +1,64 @@
...
+    layer: { id, name, description },
+    layerGroups,
+    setMapLayerGroupActive,
+    toggleMapLayerGroup
+  } = props;
+
+  const layerGroup = layerGroups.find(lg => lg.dataset === dataset.id);
+  const layerFromRedux = layerGroup && layerGroup.layers
+    && layerGroup.layers.find(l => l.id === id);
+  const layerIsActive = layerFromRedux && layerFromRedux.active;
```
tomasmoose: not super urgent but for future code monkeys it might be easier if this gets resolved in a selector, so you can return from the selector `active` and `isLayerGroup` or something similar, as those are the only conditions required and they are both booleans. My preferred solution is to have functional components as dumb as possible @pablopareja 

Example selector: https://github.com/resource-watch/resource-watch/blob/develop/components/datasets/list/selectors.js

tomasmoose:
Other than my comment above this works great!

PR #1258, opened 2020-02-24
Explore detail: save button connected to collections tooltip

pablopareja:
image

## Overview
This PR implements the feature to "save" a dataset into a collection from the new Explore detail section.
_NOTE: It also creates a new file `utils/tooltip.js` to centralize code that was repeated in different components across the app._

## Testing instructions
Go to a dataset detail's page and click on the `SAVE` button.

### DIFF
--- a/components/datasets/list/list-item/component.js
+++ b/components/datasets/list/list-item/component.js
@@ -16,6 +16,7 @@ import { breakpoints } from 'utils/responsive';
 // Tooltip
 import { Tooltip } from 'vizzuality-components';
 import CollectionsPanel from 'components/collections-panel';
+import { getTooltipContainer } from 'utils/tooltip';
 
 // helpers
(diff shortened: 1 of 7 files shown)


tomasmoose:
1. If I press <Add collection> without supplying a collection name we run into an error, we should probably block the "add collection" if no collection name is present, error bellow: 

Screenshot 2020-02-24 at 14 43 16

2. UX suggestion, when creating a new collection for the current item, it would be great if we automatically add the current one to the created one, to avoid another interaction. 

Screenshot 2020-02-24 at 14 45 35

3. Also, when a new collection is created, we should clear the collection name from the input because if I add another one the app crashes due to duplicated collection. We should probably perform a check so the user cant create a collection with the same name as an already existing one. Error is shown below: 

Screenshot 2020-02-24 at 14 48 45

PR #1259, opened 2020-02-25
Feature/add source and date

pablopareja:
imagen

## Overview
This PR adds the dataset source and `dataLastUpdated` date to the Explore detail section. Ellipsis is used for sources that don't fit in the space allocated for them.
Besides that, styles for the Explore detail sub-sections have been modified so that `row` and `column` are not used - we didn't need them for this in the end.

## Testing instructions
Check a dataset detail page such as http://localhost:9000/data/explore/for006nrt-Active-Fires-VIIRS_old_replacement
Check this dataset detail page to see the ellipsis working: http://localhost:9000/data/explore/Powerwatch
_NOTE: Bear in mind that not all datasets have this date, only or mostly the Near-real-time ones._

### DIFF
--- a/layout/explore/explore-detail/styles.scss
+++ b/layout/explore/explore-detail/styles.scss
@@ -9,28 +9,32 @@
             background-color: white;
             margin-bottom: 2 * $space;
             padding: 8 * $space 8 * $space 8 * $space 8 * $space;
-            
-            .overview {
-                .title {
-                    margin-top: $space * 4;
-                }
-                .source-date {
-                    display: flex;
-                    width: 100%;
-                    justify-content: space-between;
-                    .source {
-                        max-width: 100px;
-                        text-overflow: ellipsis;
-                    }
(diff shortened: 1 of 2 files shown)


layout/explore/explore-detail/styles.scss
```diff
@@ -5,21 +5,36 @@
...
             }
+            .source-date {
+                display: flex;
+                width: 100%;
+                justify-content: space-between;
+                margin: 2 * $space 0px 5 * $space 0px;
+                font-size: $font-size-extra-small;
+
+                .source {
+                    width: 200px;
```
tomasmoose: played around with the ellipsis and think we can allow a bit more space! super tiny detail tho

```suggestion
                    width: 80%;
```
Before:
Screenshot 2020-02-25 at 07 07 32

After: 
Screenshot 2020-02-25 at 07 07 01

PR #1269, opened 2020-03-05
Explore detail: related content section

pablopareja:
image

## Overview
This PR adds the `Related content` section to Explore detail.

## Testing instructions
Go to explore and check any given dataset.


layout/explore/explore-menu/styles.scss
```diff
@@ -4,6 +4,9 @@
     min-width: 200px;
     max-width: 200px;
     height: 100%;
+    position: sticky;
```
tomasmoose: This won't work in IE, If IE is a requirement you would need to include a polyfill https://github.com/dollarshaveclub/stickybits

pablopareja: Thanks for your input!
I didn't know that 😮 
However, we're not prioritizing having IE full support in all the app sections so maybe this is something that could be tackled in the future when necessary. 😉

PR #1270, opened 2020-03-05
Explore collections

pablopareja:
image

image

## Overview
This PR adds the user collections feature to the Explore section.

## Testing instructions
Test the interface both logged in and logged out to see the two different visual states of the component.
_NOTE: Only datasets from the collection are shown, (no widgets, dashboards, ...)_
_Also we still need to decide what to do when the user selects a collection that has no dataset included in it_.


pablopareja:
@tomasmoose I added a selector to filter out collections with zero datasets 😉

PR #1272, opened 2020-03-06
add share modal to explore detail

pablopareja:
image

## Overview
This PR adds the share functionality to the Explore detail section.

## Testing instructions
Go to any given dataset Explore detail section and click on the `SHARE` button located on the header of the section.


tomasmoose:
1. Twitter share only includes title, there is no link present to current dataset
Screenshot 2020-03-09 at 10 23 31

2. Don't have Facebook so can't test the Facebook share, but make sure there is a link present in the share dialog! 

@pablopareja

pablopareja:
@tomasmoose I've confirmed that the issue was that the bit.ly service couldn't convert a URL containing `localhost:....`. 
It works fine in the `explore` site 👇 

imagen

imagen

PR #1405, opened 2020-06-02
Very important correction

fgassert:
## Overview
This includes a very important correction to the default base map used in Pulse.

## Testing instructions
Verify that the base map looks the same but is corrected in Braddock Heights, Maryland.

## Pivotal task
NA

---


kovaca:
@pablopareja this is approved, please include in next release

PR #1591, opened 2021-02-25
adds API auth tests

andresgnlez:
## Overview
Adds API auth tests

## Testing instructions

## Pivotal task
–

---


server/auth.js
```diff
@@ -56,14 +63,36 @@ module.exports = (() => {
...
+      if (err && err.errors && err.errors[0] && err.errors[0]) {
+        const {
+          detail: errDetail,
+          status: errStatus,
+        } = err.errors[0];
+        let responseMessage = '';
+        const responseStatus = errStatus || 500;
+
+        switch (errDetail) {
+          case 'Database unavailable':
```
andresgnlez, line 75: I might remove this one.. But leaving this to give room to future error handling.

server/auth.js
```diff
@@ -56,14 +63,36 @@ module.exports = (() => {
...
+            responseMessage = errDetail || 'Something went wrong.';
+        }
+
+        return res.status(responseStatus).json({
+          status: 'error',
+          statusCode: responseStatus,
+          message: responseMessage,
+        });
       }
+      // if (!user) {
```
andresgnlez, line 91: I think we can get rid of this, as this is kind of an error and can be handled above.

server/auth.js
tiagojsag: yes.
My suggestion is doing a mapping between the response code of the API, and some user friendly error message. like, for example:
api returns 404  -> "user does not exist"
api returns 500 -> "the authentication server is temporarily unavailable, please try again later"

On future iterations you can try to map actual API responses, but this should be good enough for now, and an improvement over the current approach

server/auth.js
tiagojsag: depends on what case this covers.

test/index.spec.js
```diff
@@ -2,16 +2,99 @@ const chai = require('chai');
 const nock = require('nock');
 const { getTestServer } = require('./test-server');
 
+const userPayload = require('./payload/user');
+
 chai.should();
 
+// nock.disableNetConnect();
```
tiagojsag, line 9: once you have tests stable-ish, re-enable this and adjust as needed to make it work

PR #1602, opened 2021-03-11
runs frontend tests in production

andresgnlez:
## Overview
Please write a description. If the PR is hard to understand, provide a quick explanation of the code.

## Testing instructions

## Pivotal task
Provide the link to the task(s), if any.

---

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -71,7 +71,7 @@ const sessionOptions = {
   saveUninitialized: true,
 };
 
-if (prod) {
+if (prod && process.env.REDIS_URL) {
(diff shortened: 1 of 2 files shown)


server/app.js
```diff
@@ -71,7 +71,7 @@ const sessionOptions = {
   saveUninitialized: true,
 };
 
-if (prod) {
+if (prod && process.env.REDIS_URL) {
```
tiagojsag, line 74: redis is needed to store sessions. If it's not properly configured, we should crash the app (optionally on NODE_ENV=develop, we could just display a big red warning)

PR #1604, opened 2021-03-12
Update cypress to 6.6.0

tiagojsag:
## Overview
Please write a description. If the PR is hard to understand, provide a quick explanation of the code.

## Testing instructions

## Pivotal task
Provide the link to the task(s), if any.

---


Dockerfile.cypress
```diff
@@ -1,9 +1,9 @@
-FROM cypress/included:6.4.0
+FROM cypress/included:6.6.0
 LABEL maintainer="hello@vizzuality.com"
 
 WORKDIR /e2e
 
 RUN yarn add -D "@cypress/code-coverage@3.9.2"
-RUN yarn add -D "cypress-file-upload"
+RUN yarn add -D "cypress-file-upload@5.0.2"
```
tiagojsag, line 7: @andresgnlez as we discussed yesterday during the call: this lib is imported in 2 different places, in 2 different workflows (here and in the package.json) so we need to pin it to the exact version, to avoid differences in tests. Just an FYI for you

andresgnlez (approved):

PR #1719, opened 2021-06-30
Data/widget qc

tamaramegan:
- I have changed the structure of the widgets to meet the designs
- I have updated the widget ids to the ones in the spreadsheet

For the widgets, I have updated the staging widgets with a parameter {{geostore_env}} in the url, that will need to be changed to geostore_staging


andresgnlez (approved):
Looks good to me!

PR #1733, opened 2021-07-29
river quaility swipemap and layers

tamaramegan:
Added first swipe-map and layers to the second map


andresgnlez:
@tamaramegan the widget with id `ab52220e-768f-4d02-93d3-feca04ff9381` contains a layer (`46f41568-dd10-44e8-861a-8a552a40339b`) that it is not in staging API but production, triggering 404 in the application.

image

andresgnlez (approved):

PR #1752, opened 2021-08-17
added dynamic widgets

tamaramegan:


andresgnlez:
For the 3 widgets:
- `geostore_prod` needs  to be replaced with `geostore_id` 
- `basins.geostore_prod` should be replaced with `basins.{{geostore_env}}`

andresgnlez (changes requested):
https://github.com/resource-watch/resource-watch/pull/1752#issuecomment-900989690

PR #1755, opened 2021-08-19
Azavea development environment updates

ddohler:
## Overview
We needed to make a few small changes to the development environment to get things working on our end, primarily adjustments to how the Docker-based development setup worked. 

Additionally, this adds issue templates; this is primarily so that WRI has a template to follow when reporting issues during testing, but it can also be expanded to cover other uses later.

We've also merged our standard PR template into the existing PR template for this project. We tried to make intelligent choices about combining the two, but if something doesn't seem right, please let me know!

## Testing instructions
- I'm not sure whether the Docker development functionality is used at Vizzuality, but if you (or someone else) uses it, then it would be good to follow the Docker setup directions in the README and make sure that everything continues to work as expected.
- The other changes are just text, so reading through them to make sure they look okay should be enough.

## Jira task
N/A

## Notes
- I wasn't sure who to assign this to; if it's wrong please let me know or re-assign!
- @ethanroday you may want to take a look at the issue templates

---


andresgnlez:
Hi @ddohler, thanks for the changes! Vizzuality uses the Dockerfile to deploy the application on Jenkins and I see you have done some changes to it. I would wait for our backend, @tiagojsag (he is on holiday, back on Monday), to see if these changes can affect the current deployment process.

ddohler:
@andresgnlez -- Okay, thanks! I believe the changes should maintain existing functionality, but we should definitely have Tiago take a look to confirm that this doesn't break anything.

tiagojsag (commented):
Hey,

on the changes to the docker related files: from the description on the link in the Dockerfile, the added value of this change is that it allows easier native and containerized execution on the same Mac or Windows machine, at the expense of potentially breaking things in case a dependency assumes that node_modules is in a particular location. 

This is the first time I've seen this problem and solution (I use linux so, as the article says, this is a non-issue for me), so apologies if the following seems too conservative, but to me this has more cons than pros:
- how can we be sure that all the current deps work fine with this change?
- this effectively means a larger deviation between a native execution env and a containerized execution env, meaning devs who prefer to work natively will have to jump through additional hoops to make sure everything works fine once their code goes live.
- CI would have to be refactored to use docker too, as a "better but not perfect" way to tackle the above.

This is a personal opinion, but to me the cons outweigh the pros of this solution, and I would advise using the "keep it simple" solution listed on the same article instead. Alternatively, if you feel strongly about this particular approach, refactoring CI would be a positive step towards having more sanity now and going forward, but it's still not clear how we can be sure we're not falling into that "dependencies may break" trap

ddohler:
Hey @tiagojsag , thanks for reviewing! This issue came up during review of this change on our side too; I don't think we feel too strongly about it, but there turned out to be some additional drawbacks to the simple approach that aren't discussed in the blog post. Could you take a look at the [discussion thread on the original PR](https://github.com/azavea/resource-watch/pull/3#discussion_r658945698) and let me know what you think? I'm happy to switch to the other solution if you still prefer it. Sorry for not providing this context initially; I had forgotten how complicated it was! 🙂

tiagojsag:
that was an interesting read, hope I didn't miss any of the details. In my personal opinion (which you should feel free to ignore, in the end this is your call, I'm just advising) the best solution is to add a note to the "docker" section on the README you just added (thanks for that), saying this could happen, and how to deal with it when it does. This is basically because:
- It's as risk-free as it gets
- As the issue is present in virtually every dockerized nodejs project not actively handling it, odds are people who run into it already experienced it in the past, and know how to deal with it, even without reading the docs
- If they haven't, we're giving the their daily dose of #TIL
- I suspect the number of people who will run this code, who are on mac or windows, and who will run the code both natively and containerized, and who haven't heard about this will be very small.

PR #1770, opened 2021-08-30
Land Watch landing page

alkamin:
## Overview

This PR adds a landing page for the Land Watch content.

The page is structured to allow for content that has a maximum width as well as content that can expand to the entire width of the window.

### Notes

In order to add the dashboard to the dashboards dropdown, we'd need to add it to the production api. I have not done this.

### Screenshots/Demo

localhost_3000_dashboards_land-watch

## Testing instructions
 * ensure that your env includes `NEXT_PUBLIC_FEATURE_FLAG_LAND_WATCH=true`.
 * navigate to /dashboards/land-watch
 * it should load and look like the screenshot above
 * change the feature-flag env variable to `false`, restart the dev server
 * verify that the page is no longer visible

- [ ] Add entry to CHANGELOG.md, if appropriate


KlaasH (commented):
This looks good.  A couple copy/modify errors noted below, and a question about the title and URL.

Also, should we open a separate issue for the "The landing page should be accessible from the Dashboards menu, and should cause the Dashboard menu to highlight when users navigate to it." part?  Seems like that's more complicated than it would appear, if it's pulling the list from the production API.

ddohler:
This project is changing direction in a significant way, so we're going to close this for the time being.

PR #1843, opened 2021-09-20
Updated query for `minerals-energy` on country page

rachelthoms:
I noticed many counties are displaying 0 for this indicator because of small values. The new query should return the rounded value if it is >0, `<1` if the value is between 0 and 1, and `0` if the value is 0. Since this query returns a string, I am wondering if the percent sign should be included in the `unit` parameter or in the query itself.

## Overview

_Please write a description. If the PR is hard to understand, provide a quick explanation of the code._

## Testing instructions
 * How to test this PR
 * Prefer bulleted description
 * Start after checking out this branch
 * Include any setup required, such as bundling scripts, restarting services, etc.
 * Include test case, and expected output

## Jira task / Github issue
_Provide the link to the Jira task(s), if any._

Resolves #XXX

### Demo

Optional. Screenshots, `curl` examples, etc.

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

- [ ] Add entry to CHANGELOG.md, if appropriate


rachelthoms:
@pablopareja just flagging this PR for your review.

pablopareja (approved):

PR #1862, opened 2021-09-30
Data/hover level 5

tamaramegan:
## Overview
3 main changes:
- Added fishing widgets to home and country page
- Added logic to Mapbox styling of interactive widget to only have hover effect on costal basins. The CLICK FUNCTION is still working
- Added the layer for level 5 to the map. ATTENTION the layer has `deafult:false` and it does not render, I am not sure _where_ this should be changed

### DIFF
--- a/public/static/data/ocean-watch.json
+++ b/public/static/data/ocean-watch.json
@@ -1722,7 +1722,7 @@
                           {
                             "type": "mapnik",
                             "options": {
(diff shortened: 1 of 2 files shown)


public/static/data/ocean-watch.json
```diff
@@ -1814,7 +1811,7 @@
                   "layers": [
                     "9f857b1c-fb79-47ec-82b2-3d4788176cbf",
                     "2c5ca21a-c646-4b8c-b329-909d66ec5615",
-                    "8339b4a5-b99f-4d64-8737-30f79325e106"
+                    "1ad281f4-a6f6-4b49-b4bd-4b3087774776"
```
tamaramegan, line 1814: I tried this but didn't work
```
"layerParams":{
                    "1ad281f4-a6f6-4b49-b4bd-4b3087774776":{
                      "default":true
                      }
                    },
```

